Issue #4, opened 2016-02-01
Add error handling for missing / invalid zips


ErisDS:
In the web server version of gscan, if you hit 'Upload' without choosing a file, you get an "Internal Server Error" (or if you're in dev mode, an ugly error message and stack trace).
- [x] The form should not let you hit the button if no file is provided.

If you choose a file that's not a valid zip file, the application doesn't do anything except spin and probably eventually error out.
- [ ] The form should also validate that the file you selected is a zip file. Or show the upload form again with a simple error in red text.

PR #6, opened 2016-04-21
Disable upload button if no file is selected

acburdine:
refs #4
- watch file input and set submit button disabled property


ErisDS:
🚢 !

PR #9, opened 2016-08-15
gscan as lib, as app, as cli

kirrg001:
We want to achieve installing `gscan` via npm and use it as lib, as client or as an express app. Right now the `gscan` file structure is strongly designed to run only as express app.

- use ghost-ignition
- create app folder with all related app files for running gscan as an app
- i've tested installing this branch via npm, using the gscan cli, using gscan with require inside a node application and run the web service - worked


acburdine:
IMO we should include app.js as well. The problem with not including it in the whitelisted files is that anyone who wants to use gscan as a web application locally (to test their themes) won't be able to. Even though we don't necessarily need to use it, others might, and removing app.js means others have to clone it from github rather than doing an `npm install` and running the express server locally. The usage as an npm lib is still the same regardless of whether or not we include it.

tl;dr not including the app.js file makes the whole express part of this unusable unless cloned from github.

kirrg001:
@acburdine Thanks for feedback. We have `gscan.ghost.org` to validate a theme - no need to run a local `gscan` application for that.

acburdine:
Ok - but doesn't that run the express app?

I guess my thought is that the `files` array is for limiting which files get included in an `npm install` (mainly for filesize/amount of files issues) - and there aren't really any downsides to including it.

Even though there is `gscan.ghost.org` - I think it would be wise to let users run it locally as well. Use case: on a plane and doing theme development - can't access gscan.ghost.org but could run it locally 😄

kirrg001:
I will put all express related file into a folder called `app`. We can add it to `files`, so people can choose how they would like to use `gscan`.

Gruntfile.js
```diff
@@ -10,7 +10,7 @@ module.exports = function(grunt) {
                 options: {
                     mask: '*test.js',
                     coverageFolder: 'test/coverage',
-                    excludes: ['public', 'uploads', 'tmp', 'tpl']
+                    excludes: ['app', 'uploads', 'tmp']
                 }
             }
         },
```
kirrg001: @ErisDS 
how can i verify/ensure that the deploy still works?
I can see that `touch tmp/restart.txt` triggers a restart, but how does passenger starts the app?
`npm start`?if so, the changes should not break the deploy.

kirrg001: already discussed in slack!

PR #18, opened 2016-08-23
Find a theme anywhere in a zip, and other stories

ErisDS:
- Adds the use of glob to find `index.hbs` and count this as the base directory for the theme. If no `index.hbs` is found, we return the base directory. The checks for theme structure will output a neat error about requiring index.hbs later in the flow.

Also

- Refactors the ghost-version middleware to fail more gracefully & not stop rendering
- Added support for a bunch of helpers that were missing, else themes will be incorrectly deemed as "invalid". 
- Added some improvements to the gscan app so the website will work better. We're going to need to start using it now!


lib/read-zip.js
```diff
@@ -2,38 +2,24 @@ var debug      = require('ghost-ignition').debug('zip'),
...
-    pfs        = require('./promised-fs'),
     readZip,
     resolveBaseDir;
 
-resolveBaseDir = function readDir(zipPath, zip) {
-    zip.path = zipPath;
-    zip.origPath = zipPath;
-    zip.origName = zip.name.replace(/\.zip$/, '');
+resolveBaseDir = function resolveBaseDir(zipPath) {
+    var matches = glob.sync("**/index.hbs", {cwd: zipPath});
```
kirrg001: i would rather use the async call of `glob` if possible

ErisDS: 👍

PR #19, opened 2016-08-24
🐛 All package.json errors should be warnings

### LINKED ISSUE #15: show warning when package.json version is not valid
If the theme package.json contains e.g. `1.0`, the validation fails with an error.
We would like to change this error into a warning if possible.

ErisDS:
closes #15

- change the level from error to warning on the package.json rules
- this is wrong, but matches the current behaviour of Ghost

### DIFF
--- a/lib/spec.js
+++ b/lib/spec.js
@@ -83,7 +83,7 @@ rules = {
         "rule": "package.json file should be present."
     },
     "GS010-PJ-VAL": {
-        "level": "error",
+        "level": "warning",
         "rule": "package.json file must be valid."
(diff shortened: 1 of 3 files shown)


lib/spec.js
```diff
@@ -83,7 +83,7 @@ rules = {
         "rule": "package.json file should be present."
     },
     "GS010-PJ-VAL": {
-        "level": "error",
+        "level": "warning",
```
kirrg001: oh that was easy 👍 that looks like the real gscan way 💃

PR #23, opened 2016-08-31
fix: match partials on windows

### LINKED ISSUE #22: match partials on windows
gscan is not able to match partials on windows.
This results in an invalid theme.

kirrg001:
closes #22

gscan was not able to match partials on windows


kirrg001:
closing in favour of #24

PR #42, opened 2017-04-05
🐛  remove ignored files from the unzipped target dir

kirrg001:
no issue

- if gscan unzips the target zip file into the tmp folder and gscan detects a file which should be ignored, the file doesn't get deleted
- Ghost uses this unzipped folder for further processing
- see @TODO in the code base: there is a case, which gscan does not support right now -> remove the ignored files from the zip file (has no priority right now)

- [x] will do a quick test with Ghost


kirrg001:
Tested with Ghost 👍

PR #45, opened 2017-04-14
🐛 Fixes blocking of upload if no file or no ZIP file

### LINKED ISSUE #4: Add error handling for missing / invalid zips

Mickael-van-der-Beek:
closes #4

- button is disabled until a file has been uploaded and it's type is ZIP
- use file input tag's accept attribute to only allow uploads of files with MIME type application/file
- fallback to file type detection based on file extension for older browsers and show an error message if not a ZIP file

Note: Here's a list of browsers that don't support the `accept` attribute for `input[type=file]` tags: http://caniuse.com/#feat=input-file-accept

### DIFF
--- a/app/public/js/gscan.js
+++ b/app/public/js/gscan.js
@@ -46,11 +46,23 @@
     }
     $(document).ready(function ($) {
         if ($('#theme')[0]) {
-            $('#theme-submit').prop('disabled', !$('#theme')[0].files.length);
+            $('#theme')
+                .siblings('label')
+                .attr('for', $(this).attr('id'))
+                .toggle(false);
+            $('#theme-submit').prop('disabled', true);
         }
 
         $(document).on('change', '#theme', function () {
-            $('#theme-submit').prop('disabled', !$(this)[0].files.length);
+            var files = $(this)[0].files;
+            var hasFiles = files.length;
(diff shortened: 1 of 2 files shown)


kirrg001:
@Mickael-van-der-Beek Sorry for late review.

This PR needs a rebase and somehow when i refresh often, i see 

> File type needs to be ZIP (.zip extension).

Without doing anything.

gargol:
Additional [logic](https://github.com/TryGhost/gscan/pull/45/files#diff-c4cbaf96b0c41c320ae17671c95e166eR49) in gscan.js file is causing disabled submit button when navigating back from the check page trying to change the version. Think that just adding the `accept` attribute to the file input field should be enough of the improvement on the frontennd side and some basic type checking on the server side should cover the rest of cases. Closing this PR as it's very out of date at this point.

PR #50, opened 2017-05-30
package.json validation && fixture cleanup

kirrg001:
**WIP!**

- [x] posts per page
- [x] do not show package.json errors when using Ghost production env - only on gscan.ghost.org and for Ghost development
- [x] code quality check && more testing
- [x] add more context to the commits (e.g. why PJV was removed)

Fatal errors are chosen via a different issue, see https://github.com/TryGhost/Ghost/issues/8222. The idea behind it is, to select a couple of errors, which are fatal (you can't activate this theme in Ghost). And the other errors are still errors, but not fatal. We will add a little attribute for this.

I've already tested some real themes, looks good so far (for gscan.ghost.org). But need's more testing tomorrow.

I went a bit crazy with the theme fixtures, but it annoyed me very hard!


AileenCGN:
Tested with different themes and errors in `package.json` LGTM 👍

PR #53, opened 2017-06-08
🐛  Improve deprecation RegEx

AileenCGN:
closes TryGhost/Ghost#8543

Fixes a bug, where the regex for deprecation checks was wrong because of the usage of `.` in the RegEx rather than `\.` 😔

### DIFF
--- a/lib/checks/001-deprecations.js
+++ b/lib/checks/001-deprecations.js
@@ -6,69 +6,69 @@ var _       = require('lodash'),
 checkDeprecations = function checkDeprecations(theme, themePath) {
     var checks = [
             {
-                helperRegEx: new RegExp(/{{pageUrl(\s\S+)?}}/i),
+                helperRegEx: /{{pageUrl(\s\S+)?}}/ig,
                 helperName: '{{pageUrl}}',
                 ruleCode: 'GS001-DEPR-PURL'
             },
             {
-                helperRegEx: new RegExp(/<meta name=("|')description("|') content=("|'){{meta_description}}("|')/i),
+                helperRegEx: /<meta name=("|')description("|') content=("|'){{meta_description}}("|')/ig,
                 helperName: '{{meta_description}}',
                 ruleCode: 'GS001-DEPR-MD'
             },
             {
-                helperRegEx: '{{image}}',
+                helperRegEx: /{{image}}/g,
                 helperName: '{{image}}',
                 ruleCode: 'GS001-DEPR-IMG'
             },
             {
-                helperRegEx: '{{author.image}}',
+                helperRegEx: /{{author\.image}}/g,
                 helperName: '{{author.image}}',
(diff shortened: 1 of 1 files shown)


AileenCGN:
So, here my two thoughts on this:

This fix _would_ detect the usage of `{{author.image}}` in the exact way, *and only that*. The usage of `{{author_image}}` which is not correct as well (should be `{{author.profile_image}}`), stays then undetected.

If we leave gscan the way it is now, it _would_ detect this false usage, but the message is  confusing, as it explicitly mentions the usage of a detected `{{author.image}}`. 

Would like to here your thoughts about that, @JohnONolan && @kevinansfield !

lib/checks/001-deprecations.js
```diff
@@ -21,32 +21,32 @@ checkDeprecations = function checkDeprecations(theme, themePath) {
                 ruleCode: 'GS001-DEPR-IMG'
             },
             {
-                helperRegEx: '{{author.image}}',
+                helperRegEx: new RegExp(/{{author\.image}}/),
```
ErisDS: Escaping the dot will fix the issue :)

Consistency Notes:

- the use of `new RegExp()` isn't really needed, it's useful if you need to build a dynamic regex, e.g. you want to concatenate a variable in like:
`var myRegex = new Regex(myVar1+ myVar1)`
However here I don't think we ever really need it as regexes can be defined as `helperRegex: /{{post\.image}}/`
- Some regexes have the `g` flag, some don't - I think this should be consistent. I believe the `g` flag will make the rule ind ALL instances of the match, and without it will only match the first.
- If you are using `new RegExp()` flags should be passed as the second argument, E.g. `new RegExp(/{{post\.image}}/, 'g')` otherwise it would be `helperRegex: /{{post\.image}}/g`
- `{` is also a special char in regexes, I was going to suggest escaping the first one only to be clear this is not meant to be a special char but testing with https://regex101.com/ shows this is really not necessary.

TLDR; this will fix the bug, there are also some inconsistencies.

ErisDS:
IMO the purpose of the deprecation check is to find themes which are specifically using the old syntax, and tell them about the new syntax. 

Looking for typos or other issues where theme devs have made a mistake is a different problem, which we could look to solve but I think it would be quite hard!

lib/checks/001-deprecations.js
AileenCGN: Awesome! Thanks a lot! For the sake of consistency, I changed it now, so every regex has the `g` flag.

kirrg001:
@AileenCGN Could you add 1-2 tests after the beta? Thanks 👍

AileenCGN:
Will do!

AileenCGN:
@kirrg001 I added some more test cases in the theme fixtures via #55. Can you check if this is enough?

PR #55, opened 2017-06-12
🐷  Add more test cases for deprecation usage

AileenCGN:
no issue

Adds wrong usage cases for deprecation checks to prove that they won't be detected and therefore don't cause errors.


AileenCGN:
Follow-up PR for #53

PR #57, opened 2017-06-19
Fix extension parsing

sebgie:
Extension parsing doesn’t use the last part of the extension but
everything that follows after a dot.

Example - logo.new.hbs
- old: .new.hbs
- new: .hbs

Changes:
- changed to only use numbers and characters after a dot as file
extension
- added `i` to make regex case insensitive
- added test

### DIFF
--- a/test/general.test.js
+++ b/test/general.test.js
@@ -167,15 +167,21 @@ describe('Read theme', function () {
         readTheme(themePath('partials')).then(function (theme) {
             theme.should.be.a.ValidThemeObject();
 
-            theme.files.should.be.an.Array().with.lengthOf(6);
-            var fileNames = _.map(theme.files, 'file');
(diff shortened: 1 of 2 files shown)


kirrg001:
Added to the Ghost sprint 👍 Review will happen before Thursday.

test/general.test.js
```diff
@@ -167,15 +167,21 @@ describe('Read theme', function () {
         readTheme(themePath('partials')).then(function (theme) {
```
kirrg001: This test read's a theme which is called `partials`, which contains a folder `partials`. Somehow confusing.
I know this is not part of your task here, but could you please rename this folder to e.g. `theme-with-partials`?

sebgie:
Added a second commit to rename the theme :+1:.

PR #58, opened 2017-06-19
Ignore `.svn` folders

sebgie:
- added `.svn` to the ignore list (same as `.git`)


kirrg001:
Added to the Ghost sprint 👍 Review will happen before Thursday.

AileenCGN:
LGTM 👍

PR #59, opened 2017-06-19
Regex hangs

sebgie:
The regex to filter hbs files is written in a way that it can create a
condition where gscan will hang forever. This is similar to an endless
loop and not handled with an exception in JS Regex. This condition is
also referred to as „catastrophic backtrack“.

Pre Condition:
- hbs and a second extension (e.g.: .hbs.old)
- part of the partials folder
- relatively long filename/filepath (> 35 chars)

Changes:
- changed/optimized regex
- added test file to reproduce the case


AileenCGN:
Tested without the changes and can confirm this fix! LGTM 👍

PR #65, opened 2017-06-27
🐛 Fix partial paths on windows

### LINKED ISSUE #64: Path issue with partials on Windows
A user got this error when trying to start Ghost Beta.2

image

The error appears to be a problem with paths to partials on windows.

- reference partial is `partials\\floating-header.hbs` (not sure why there are 2 slashes)
- error message talks about the partial `icons/point` (slash is forward... Windows not like this)

I am not 100% if this is a gscan bug, but as Casper 2.0 is working in all of my tests on mac&linux it would make the most sense that there is some missing place in gscan where we need to add some extra windows handling.

I'm pretty sure that nested partials works on windows normally...

Steps to reproduce:
- use windows
- install latest release from the zip
- start ghost

OS: Windows 10.0.15063 x64

ErisDS:
Note, I don't have windows to test this on, but the user who reported the issue in Slack tested it out and it worked ([see](https://ghost.slack.com/archives/C04CURXK7/p1498589498373044)).

fixes #64 

- ref: https://github.com/barc/express-hbs/issues/84
- express-bhs does this already for us in Ghost
- we need to do it here to ensure that the read path (which will be back-slashed on windows) matches
  the partial path in the template, which should always use forward slashes.


ErisDS:
I realised that this commit and #69 will interact / conflict. Seeing as @cobbspur you're the person with a working windows setup (I think) and you also did #69 mayhaps you can take this over?

Happy for you to apply the same fix in #69 and just have one PR - don't have to fuss around with 2.

Issue #63, opened 2017-06-27
Unkown helper not always detected when within context


AileenCGN:
I've stumbled over an inconsistency. While intentionally using unknown hbs helpers in themes, I found that they sometimes get detected and sometimes not. The reason for that is the context they are in. A helper that is definitely wrong e. g. `{{bla "string"}}` is not detected within e. g. `author` context.

---

### Example that works fine in `author.hbs`:

![image](https://user-images.githubusercontent.com/8037602/27617424-6895c9f4-5be0-11e7-8ef8-658fa551cc44.png)
*Note, that this ^ is within the `author` context!!*

#### Example theme:
[casper.zip](https://github.com/TryGhost/gscan/files/1104961/casper.zip)

#### Output:

*Output, when outside of `author` context*:
![image](https://user-images.githubusercontent.com/8037602/27617541-35ee88c8-5be1-11e7-9ac2-df7ac9417e06.png)

*Output when inside of `author` context*:
![image](https://user-images.githubusercontent.com/8037602/27617674-f75ddf90-5be1-11e7-9521-c4fec66986b2.png)

---

### Example, where it fails, detecting wrong helpers in `author.hbs`:

![image](https://user-images.githubusercontent.com/8037602/27617492-dc7aee30-5be0-11e7-8ecd-d62353e215c7.png)
*Note, that this ^ is outside of the `author` context.

#### Example theme:
[Ghostly-master 2.zip](https://github.com/TryGhost/gscan/files/1104963/Ghostly-master.2.zip)

#### Output:

*Fails to detect wrong helper, when within `author` context*:
![image](https://user-images.githubusercontent.com/8037602/27617509-fa92ee68-5be0-11e7-8401-34f3542c73b2.png)

*Detection works in this theme, when outside of `author` context*:
![image](https://user-images.githubusercontent.com/8037602/27617610-b0ee167e-5be1-11e7-9ce1-c9e21cbf109c.png)

ErisDS:
@AileenCGN can you please provide the example code that generates an unexpected output, and also what the output is, and what is unexpected about it?

AileenCGN:
@ErisDS Thanks for the suggestion! I updated the issue description and provided more information.

cobbspur:
This is because no data is being passed to the template, therefore blocks like author are not executed.

Also partials content is not added when the partial is registered.

PR #68, opened 2017-07-05
Make image helper errors fatal 

### LINKED ISSUE #67: Old {{image}} helper is not being detected within a partial file.
I was updating a old theme so that it can be compatible with Ghost 1.0.0. I have found this issue. If it's intended behavior then feel free to close this issue.

### Exact case

I have `{{tag.image}}` within a partial file named **cover-tag.hbs** . Which is being called within **tag.hbs** file.

As we know, In ghost 1.0.0 `{{image}}` or in this case `{{tag/image}}` can not be used. So gscan should show error or warning. But If I upload the theme including this code there is no error or warning. Theme check passed.

**Code in the cover-tag.hbs file**

```
<div class="row cover tag-cover {{#if tag.image}}" style="background-image: url({{tag.image}}) {{else}}no-image{{/if}}">
	<div class="col-md-12 cover-inner">
		<h3 class="tag-name">
			Tag: {{tag.name}}
		</h3>
		<div class="post-count">
			Total {{plural pagination.total empty='0 Post' singular='% Post' plural='% Posts'}}
		</div>
		{{#if tag.description}}
		<div class="tag-description">
	        {{tag.description}}
		</div>
		{{/if}}
	</div>
</div>
```

**Code in tag.hbs file where cover-tag.hbs file is being called**

```

{{!< default}}
{{! The tag above means - insert everything in this file into the {body} of the default.hbs template }}

<div class="content-area-wrap">
	{{> mobile-header}}
	<!-- start main content -->
	<div class="container-fluid main-content-wrap">
		{{> cover-tag}}
		<div class="row posts-wrap-outer">
			{{#foreach posts}}
			<!-- start post -->
			.............................................................................................................
			...........this part of code is not relevant for this issue ...........................
			.............................................................................................................
			<!-- end-post -->
			{{/foreach}}
		</div>
		{{pagination}}
	</div>
	<!-- end main content -->
</div>
```

**Passed when uploading in to Ghost dashboard** 

after upload

**Passed when uploading in to gscan.ghost.com** 

theme check passed

cobbspur:
Attempted to cover the most common use cases of image helpers 

refs #63, closes #67

- Changes errors relating to image helper to fatal errors.
- Allows whitespace and line breaks to be included in checks for deprecated handlebars.
- Fixes regex for tag.image
- Add common nested image regex 
- Add regex checks for image conditionals


lib/checks/001-deprecations.js
```diff
@@ -5,75 +5,125 @@ var _       = require('lodash'),
...
+            helperName: '{{author.cover}}',
+            ruleCode: 'GS001-DEPR-AC'
+        },
+        {
+            helperRegEx: /{{[\s\S]?tag\.image[\s\S]?}}/g,
+            helperName: '{{tag.image}}',
+            ruleCode: 'GS001-DEPR-TIMG'
+        },
+        {
+            helperRegEx: /{{[\s\S]?post.author\.image[\s\S]?}}/g,
```
kirrg001: An escape of the dot is missing. Otherwise it would match any character.

cobbspur:
@kirrg001 passing now. I think the only other case I haven't added yet is {{ tags.[4].image}} from say a post context. I can that and a corresponding conditional in if the rest of this looks sane

test/fixtures/themes/001-deprecations/invalid/post.hbs
```diff
@@ -1,6 +1,30 @@
...
+    {{post.author.cover}}
+{{/if}}
+
+{{#if post.tags.[3].image}}
+    {{post.tags.[3].image}}
+{{/if}}
 {{#post}}
     {{meta_description}}
-    {{image}}
+    {{#if image}}
```
kirrg001: How about adding one example with white spaces?

lib/checks/001-deprecations.js
```diff
@@ -5,75 +5,125 @@ var _       = require('lodash'),
...
-            },
-            {
-                cssRegEx: /\.page-template-\w+[\s\{]/g,
-                className: '.page-template-slug',
-                css: true,
-                ruleCode: 'GS001-DEPR-CSS-PATS'
-            }
-        ];
+        {
+            helperRegEx: /{{\s*?pageUrl['"\w\d\s]*?}}/ig,
```
kirrg001: This regex also matches {{pageUrlA]]?
But it should only match [[pageUrl something]].

PR #69, opened 2017-07-08
Ensure partial content is compiled

cobbspur:
refs #63
- find and insert content from partials content when registering in handlebars


cobbspur:
If you can have sub-partials this PR won't quite work, will need a small tweak I imagine.

kirrg001:
@cobbspur Are you pushing a fix for sub partials here?

PR #70, opened 2017-07-08
Provide fake data to hbs compiler

### LINKED ISSUE #63: Unkown helper not always detected when within context

cobbspur:
closes #63
- provides fake data to allow author/post/tag blocks to execute
- uses handlebars {{each}} helper to replace {{foreach}} helper

### DIFF
--- a/lib/checks/005-template-compile.js
+++ b/lib/checks/005-template-compile.js
@@ -7,7 +7,8 @@ var _       = require('lodash'),
 checkTemplatesCompile = function checkTemplatesCompile(theme) {
     var ruleCode = 'GS005-TPL-ERR',
         localHbs = hbs.create(),
-        failures = [];
+        failures = [],
+        missingHelper;
 
     _.each(theme.partials, function (partial) {
         var partialContent = _.find(theme.files, {file: 'partials/' + partial + '.hbs'}).content;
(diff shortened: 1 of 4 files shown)


cobbspur:
Ahh nice. I would still like to keep the fake data in there. It something we can look to build on later to determine context should we wish, maybe we can add an option to override the fakeData, for tests etc

kirrg001:
> Ahh nice. I would still like to keep the fake data in there

yeah i kept them 👍

PR #72, opened 2017-07-13
👀  Scan for `{{cover}}` change `{{image}}` wording

AileenCGN:
no issue

- Scans for the usage of `{{cover}}` with advice to change it to `{{cover_image}}`.
- Changes wording of `{{image}}` advice to not include `{{cover_image}}`

TODO:
- [x] manual tests

### DIFF
--- a/lib/checks/001-deprecations.js
+++ b/lib/checks/001-deprecations.js
@@ -21,7 +21,7 @@ checkDeprecations = function checkDeprecations(theme, themePath) {
             ruleCode: 'GS001-DEPR-IMG'
         },
         {
-            helperRegEx: /{{\s*?cover\b[\w\s='"]*?}}/g,
+            helperRegEx: /{{\s*?cover\s*?}}/g,
             helperName: '{{cover}}',
             ruleCode: 'GS001-DEPR-COV'
         },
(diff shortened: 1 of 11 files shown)


AileenCGN:
Successfully detects `{{cover}}` usage:

image

cobbspur:
We need to ensure that our recommendations results in matching behaviour.

`{{feature_image}}` etc are purely data attributes not helpers in 1.0, therefore `{{image}}` would need to change to `{{img_url "feature_image"}}`

lib/checks/001-deprecations.js
```diff
@@ -21,6 +21,11 @@ checkDeprecations = function checkDeprecations(theme, themePath) {
             ruleCode: 'GS001-DEPR-IMG'
         },
         {
+            helperRegEx: /{{\s*?cover\b[\w\s='"]*?}}/g,
```
cobbspur: This is just a data attribute ( there won't won't be any valid text like absolute='true') therefore we can follow the format `\s*?` instead of `\b[\w\s='"]*?`

cobbspur: Also we want to have a conditional check as themes are quite likely to use `{{# if cover}}` e.g `/{{\s*?#if\s*?cover\s*?}}/g`

AileenCGN:
@cobbspur done! 🎉

cobbspur:
This line  https://github.com/TryGhost/gscan/pull/72/files#diff-2a21f9253b13198947b679b60b584c0bR89 still uses {{feature_image}}, did you see my previous comment 

> {{feature_image}} etc are purely data attributes not helpers in 1.0, therefore {{image}} would need to change to {{img_url "feature_image"}}

AileenCGN:
Bummer... Yup... Didn't see this line. I changed it. Hopefully all good now 😬

PR #74, opened 2017-07-14
📖  Error details in GScan

AileenCGN:
no issue

Adds details for each rule incl. links to documentations and examples for the correct usage.

Todos:
- [x] proofread!! (wording and links)
- [x] check how this behaves with Ghost Admin 🙈 -> in progress, see https://github.com/TryGhost/Ghost-Admin/pull/784
- [x] use `<code>` wherever possible

### DIFF
--- a/lib/spec.js
+++ b/lib/spec.js
@@ -77,211 +77,342 @@ templates = [
 rules = {
     "GS001-DEPR-PURL": {
         "level": "error",
-        "rule": "Replace <code>{{pageUrl}}</code> with <code>{{page_url}}</code>.",
-        "fatal": true
+        "rule": "Replace <code>{{pageUrl}}</code> with <code>{{page_url}}</code>",
+        "fatal": true,
+        "details": "The helper <code>{{pageUrl}}</code> was replaced with <code>{{page_url}}</code>.<br>" +
+                   "Find more information about the <code>{{page_url}}</code> helper <a href=\"https://themes.ghost.org/docs/page_url\" target=_blank>here</a>."
     },
     "GS001-DEPR-MD": {
         "level": "error",
-        "rule": "Usage of <code>{{meta_description}}</code> in HTML <code>head</code> is deprecated."
+        "rule": "The usage of <code>{{meta_description}}</code> in HTML <code>head</code> is no longer required",
+        "details": "The usage of <code>{{meta_description}}</code> in the HTML <code>head</code> tag is no longer required because Ghost outputs this for you automatically in <code>{{ghost_head}}</code>.<br>" +
+                   "Check out the documentation for <code>{{meta_description}}</code> <a href=\"https://themes.ghost.org/docs/meta_description\" target=_blank>here</a>.<br>" +
+                   "To see, what else is rendered with the <code>{{ghost_head}}</code> helper, look <a href=\"https://themes.ghost.org/docs/ghost_head\" target=_blank>here</a>."
     },
     "GS001-DEPR-IMG": {
         "level": "error",
-        "rule": "Replace the <code>{{image}}</code> helper with <code>{{img_url \"feature_image\"}}</code> or <code>{{img_url \"profile_image\"}}</code>.",
-        "fatal": true
+        "rule": "Replace the <code>{{image}}</code> helper with <code>{{img_url feature_image}}</code> or <code>{{img_url profile_image}}</code>",
+        "fatal": true,
+        "details": "The <code>{{image}}</code> helper was replaced with <code>{{img_url}}</code>.<br>" +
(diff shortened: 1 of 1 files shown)


AileenCGN:
This is how it looks like (examples):
image

ErisDS:
This is great 👍 

I have a few wording tweaks:

Little thing: can we change "and got replaced with" -> "and was replaced with".

Also "deprecated" has always meant for us, we recommend you don't use it any more but you can (e.g. pageUrl for the past year or two). Whereas the features have now been fully removed. Therefore I prefer the instances which say:

> "The xxx attribute was replaced with xx

And think we should also use that pattern for "The xxx _helper_ was replaced with xx"

Similarly, for meta_description I would change:

> The usage of <code>{{meta_description}}</code> in the HTML \"head\" tag is deprecated

To something like

> The usage of <code>{{meta_description}}</code> in the HTML \"head\" tag is no longer required

and the longer form one

> The usage of <code>{{meta_description}}</code> in the HTML \"head\" tag is no longer required because Ghost outputs this for you automatically in `{{ghost_head}}`.

lib/spec.js
```diff
@@ -78,210 +78,338 @@ rules = {
...
         "level": "error",
-        "rule": "package.json <code>author.email</code> must be valid."
+        "rule": "<code>package.json</code> property <code>\"author.email\"</code> must be valid.",
+        "details": "The property <code>\"author.email\"</code> in your <code>package.json</code> file must a valid email. E.g. <code>{\"author\": {\"email\": \"hello@example.com\"}}</code>.<br>" +
+                   "Check the <a href=\"https://themes.ghost.org/docs/packagejson#section--author-\" target=_blank><code>package.json</code> documentation</a> for further information."
     },
     "GS010-PJ-CONF-PPP": {
         "level": "recommendation",
-        "rule": "package.json <code>config.page_per_post</code> is recommended. Otherwise, it falls back to 5."
+        "rule": "<code>package.json</code> property <code>\"config.page_per_post\"</code> is recommended. Otherwise, it falls back to 5.",
```
ErisDS: Just found an existing bug - `page_per_post` is the wrong name. Should be `posts_per_page`

lib/spec.js
```diff
@@ -78,210 +78,338 @@ rules = {
...
+        "rule": "<code>package.json</code> property <code>\"config.page_per_post\"</code> must be integer.",
+        "details": "The property <code>\"config.page_per_post\"</code> in your <code>package.json</code> file must be integer. E.g. <code>{\"config\": { \"posts_per_page\": 5}}</code>.<br>" +
+                   "Check the <a href=\"https://themes.ghost.org/docs/packagejson#section--config-posts_per_page-\" target=_blank><code>package.json</code> documentation</a> for further information."
     },
     "GS010-PJ-AUT-EM-REQ": {
         "level": "error",
-        "rule": "package.json <code>author.email</code> is required."
+        "rule": "<code>package.json</code> property <code>\"author.email\"</code> is required.",
+        "details": "Please add the property <code>\"author.email\"</code> to your <code>package.json</code>. E.g. <code>{\"author\": {\"email\": \"hello@example.com\"}}</code>.<br>" +
+                   "Check the <a href=\"https://themes.ghost.org/docs/packagejson\" target=_blank><code>package.json</code> documentation</a> to see which properties are required and which are recommended."
```
ErisDS: Possibly worth explaining about this property?

Basically, it is required so that themes which are distributed (either free or paid) have a method of contacting the author so users can get support and more importantly so that we can reach out about breaking changes and security updates.

Also worth explaining that the package.json file is NOT accessible when uploaded to a blog so if the theme is only uploaded to a single blog, no one will see this email address.

ErisDS:
This is looking good 👍

lib/spec.js
```diff
@@ -78,210 +78,338 @@ rules = {
...
+                   "Check out the documentation for <code>{{meta_description}}</code> <a href=\"https://themes.ghost.org/docs/meta_description\" target=_blank>here</a>.<br>" +
+                   "To see, what else is rendered with the <code>{{ghost_head}}</code> helper, look <a href=\"https://themes.ghost.org/docs/ghost_head\" target=_blank>here</a>."
     },
     "GS001-DEPR-IMG": {
         "level": "error",
         "rule": "Replace the <code>{{image}}</code> helper with <code>{{img_url \"feature_image\"}}</code> or <code>{{img_url \"profile_image\"}}</code>.",
-        "fatal": true
+        "fatal": true,
+        "details": "The <code>{{image}}</code> helper was replaced with <code>{{img_url}}</code>.<br>" +
+                   "Depending on the context of the <code>{{img_url}}</code> helper you would need to use e. g. <br><br><code>{{#post}}<br>&nbsp;&nbsp;&nbsp;&nbsp;{{img_url \"feature_image\"}}<br>{{/post}}</code><br><br>to render the feature image of the blog post.<br>" +
```
ErisDS: Also there should be no quotes around the property passed to `{{img_url}}`

lib/spec.js
AileenCGN: Good catch!!!

PR #76, opened 2017-07-26
🚨 Add checks for {{id}} and disqus embeds

ErisDS:
See https://github.com/TryGhost/Ghost/issues/8760 for the reasoning behind these additional checks.

Here's my regex101 test ground for the  regex which looks for the Disqus embed: https://regex101.com/r/AFj5Ml/3

- Ghost 1.0 changed id format from integers to ObjectID strings
- This breaks disqus comments 😞
- We're rolling out an emergency workaround to use `{{comment_id}}` instead
- This will contain the old id for posts that get migrated from LTS
- Anyone using the disqus embed MUST update their theme in order to keep comments
- Anyone using `{{id}}` MAY also want to switch over to the new property

### DIFF
--- a/lib/checks/002-comment-id.js
+++ b/lib/checks/002-comment-id.js
@@ -1,0 +1,48 @@
+var _ = require('lodash'),
+    checkCommentID;
+
+checkCommentID = function checkCommentID(theme, themePath) {
+    var checks = [
+        // Check 1:
+        // Look for instances of: this.page.identifier = 'ghost-{{id}}';
+        // Also the old style disqus embed: var disqus_identifier = 'ghost-{{id}}';
+        {
+            regex: /(page\.|disqus_)identifier\s?=\s?['"].*?({{\s*?id\s*?}}).*?['"];?/g,
+            ruleCode: 'GS002-DISQUS-ID'
+        },
+        // Check 2:
+        // Look for other usages of {{id}}
+        {
+            regex: /({{\s*?id\s*?}})/g,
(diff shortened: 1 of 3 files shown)


ErisDS:
Note: I have updated the regex to handle both the modern style of disqus embed:

Modern:
`this.page.identifier = 'ghost-{{id}}';`
Old:
`var disqus_identifier = 'ghost-{{id}}';`

lib/checks/002-comment-id.js
```diff
@@ -0,0 +1,48 @@
...
+    var checks = [
+        // Check 1:
+        // Look for instances of: this.page.identifier = 'ghost-{{id}}';
+        // Also the old style disqus embed: var disqus_identifier = 'ghost-{{id}}';
+        {
+            regex: /(page\.|disqus_)identifier\s?=\s?['"].*?({{\s*?id\s*?}}).*?['"];?/g,
+            ruleCode: 'GS002-DISQUS-ID'
+        },
+        // Check 2:
+        // Look for other usages of {{id}}
```
kirrg001: What are these other usages? 😬

ErisDS:
We have a fatal error for anything that is _clearly_ a disqus embed, and a warning for anything else.

Other cases where `{{id}}` might have been used is in get helper filters (doesn't need updating) or classnames (probably does need updating) therefore this is a warning rather than an error to let the theme owner decide.

ErisDS:
Updated to include a reference to https://github.com/TryGhost/Ghost/issues/8760

PR #87, opened 2017-09-27
✨ Extract custom templates 

kirrg001:
refs TryGhost/Ghost#9060

- extract `templates` object
  - all templates
  - custom templates


test/general.test.js
```diff
@@ -186,6 +186,38 @@ describe('Read theme', function () {
             done();
         });
     });
+
+    it('Can extract custom templates', function (done) {
```
kirrg001: @kevinansfield If you can think of more test cases (e.g. more example file names), let me know 👍

lib/read-theme.js
```diff
@@ -1,21 +1,24 @@
 var Promise = require('bluebird'),
-    _       = require('lodash'),
-    os     = require('os'),
-    hbs     = require('express-hbs'),
-    path    = require('path'),
-    pfs     = require('./promised-fs'),
+    _ = require('lodash'),
```
kirrg001: @kevinansfield You can ignore the changes till `extractCustomTemplates` starts, just indentions and lint errors.

lib/read-theme.js
```diff
@@ -105,16 +108,84 @@ processHelpers = function (theme, themeFile) {
...
+module.exports = function readTheme(themePath) {
+    return readThemeStructure(themePath)
+        .then(function (themeFiles) {
+            var allTemplates = extractTemplates(themeFiles);
+
+            return readHbsFiles({
+                path: themePath,
+                files: themeFiles,
+                templates: {
+                    all: allTemplates,
```
kirrg001: See https://github.com/TryGhost/Ghost/blob/master/core/server/themes/active.js#L45
It's useful for Ghost to return both: all templates and the custom templates.

test/general.test.js
kevinansfield: Only thing I can think of testing would be nested templates, eg `post-partials/footer.hbs`.

I know the `(^[^\/]+).hbs$` regex protects against those but it seems like one of those things that could regress if we allow deeper nesting at some point.

kirrg001: I can add a test to show that if you add a `post-partials` folder, it won't be counted as a custom template. Is this what you mean? 🙈

kevinansfield: Yeah, I was just thinking of adding a `post-partials/footer.hbs` template to the test theme. The length checks should already act as a test case for it not being included?

kirrg001: Okay can do, will add as second commit.

kirrg001: https://github.com/TryGhost/gscan/pull/87/commits/59d66b1a1b24a4071608c36b9db5dc8561e2f872

PR #91, opened 2017-11-17
Allow custom helpers in env variable

benaubin:
This allows users to create custom apps that specify templates and allow them through ENV variables. This makes it hard to activate, but possible for developers (so only the people who should be using custom helpers use custom helpers).

### DIFF
--- a/lib/spec.js
+++ b/lib/spec.js
@@ -18,7 +18,7 @@ knownHelpers = [
     'log', 'if', 'unless', 'with', 'block', 'contentFor', 'each', 'lookup'
     // Registering these will break template compile checks
     // 'blockHelperMissing', 'helperMissing',
-];
+].concat((process.env.GSCAN_ALLOW_HELPERS || "").split(","));
 
 templates = [
     {


ertrzyiks:
Great! That would unblock me from migrating to ghost 1.x. This or an option like

`I understand the risk, activate theme them anyway`

benaubin:
Yeah, this was just a quick & easy fix. I would prefer a understand the risks prompt, but, In the meantime, you can monkey patch this in `current/node_modules/gscan/lib/spec.js`.

ErisDS:
I fully understand the problem here and why a way to do this is useful, however, I'm not 100% on the approach to a solution in this PR. I need to think about this a little bit from the perspective of what are apps likely to look like when they finally happen, aaaaand how will they interact with gscan. The end goal being to figure out if there's a obvious future-proof path here. 

Let me have a think and come back to you. Ultimately, custom helpers _are_ going to be possible, just need to figure out how.

benaubin:
Cool! I totally agree! The goal of this PR is mainly just a way to give developers a temporary solution - not to create a future-proof one. The nice thing about ENV variables is that they can be accessed from any part of the app, so this solution is something that can be future proof, but it also allows implementations of gscan that aren't running with apps to support custom helpers. I think app-wide solutions will be better, and what I'd prefer, but I think gscan will need to accept the custom helpers via env variable (or possibly arguments, but ENV variables support different use cases, maybe both?) when running outside of ghost.

kirrg001:
Sorry for late response. I'll close this pull request, because apps are tracked on our internal board and we would like to come up with a full solution to this problem. I hope you can understand our decision.

If you would like to suggest a new helper, which might be useful for many developers, let us know in our [forum](https://forum.ghost.org). We want to hear about use case and are open to expand the core helpers.

PR #98, opened 2018-03-26
use paths within repo to test symlinks

### LINKED ISSUE #97: links to /etc/passwd and /tmp
Why are there symlinks to `/etc/passwd` and `/tmp` in the repo?  This is not safe.

richRemer:
fixes #97


kirrg001:
You've added a new image foo.png, which doesn't load here in Github? 🤔

richRemer:
It's simply a file for the symlink to target.  Its name and content are irrelevant.

PR #103, opened 2018-04-16
🎨 Switched to eslint plugin

AileenCGN:
no issue

Use our standardised eslint plugin and fix linting issues.

The commits can me squashed. I only left them separate for readability reasons.


shipitfile.js
```diff
@@ -6,14 +6,14 @@ function init(shipit) {
             yarn: true,
             workspace: './',
             deployTo: '/opt/gscan/',
-            ignores: ['.git', '.gitkeep', '.gitignore', '.jshintrc', 'node_modules'],
+            ignores: ['.git', '.gitkeep', '.gitignore', '.jshintrc', 'node_modules']
```
kirrg001: > '.jshintrc'

AileenCGN: 👍 Fixed and pushed!

app/index.js
```diff
@@ -76,9 +75,9 @@ app.use(function (req, res, next) {
     next(new errors.NotFoundError({message: 'Page not found'}));
 });
 
-app.use(function (err, req, res, next) {
+app.use(function (err, req, res) {
```
kirrg001: This doesn't work anymore.
See for example: https://github.com/expressjs/express/issues/2896

lib/read-theme.js
```diff
@@ -10,15 +10,17 @@ var Promise = require('bluebird'),
...
 readThemeStructure = function readThemeFiles(themePath, subPath, arr) {
     themePath = path.join(themePath, '.');
     subPath = subPath || '';
 
     var tmpPath = os.tmpdir(),
-        inTmp = themePath.substr(0, tmpPath.length) === tmpPath,
-        arr = arr || [];
+        inTmp = themePath.substr(0, tmpPath.length) === tmpPath;
+
+    arr = arr || [];
```
kirrg001: where is the variable definition

shipitfile.js
```diff
@@ -6,14 +6,14 @@ function init(shipit) {
             yarn: true,
             workspace: './',
             deployTo: '/opt/gscan/',
-            ignores: ['.git', '.gitkeep', '.gitignore', '.jshintrc', 'node_modules'],
+            ignores: ['.git', '.gitkeep', '.gitignore', 'node_modules']
```
kirrg001: you need to add the eslint files?

lib/read-theme.js
AileenCGN: It's passed to the function?

shipitfile.js
AileenCGN: 🤦‍♀️

lib/read-theme.js
kirrg001: ok i see 👍

app/index.js
AileenCGN: Good call! Didn't know that!

AileenCGN:
@kirrg001 This should be good to go now

PR #107, opened 2018-05-11
Added error handling for zip files

### LINKED ISSUE #106: Improve error handling for reading zip files
refs TryGhost/Support#426

GScan can crash (and with it Ghost), when a corrupted zip file is uploaded.

The error happens in a subdependency ([yauzl](https://github.com/thejoshwolfe/yauzl/blob/master/index.js#L227)) where an `error` event is fired, but not handled in the dependency we're using: [extract-zip](https://github.com/maxogden/extract-zip)

There's is already an open [issue](https://github.com/maxogden/extract-zip/issues/65) and [PR](https://github.com/maxogden/extract-zip/pull/63) to fix this, but no communication about it yet.

Nevertheless, GScan still needs a better way to handle this situation.

AileenCGN:
closes #106

GScan didn't handle errors `.zip` file errors fully. Furthermore, there is a bug in one of the underlying dependencies, which does not handle and return an error event correctly. Until https://github.com/maxogden/extract-zip/issues/65 is fixed, we therefore need to use our forked version, which includes the fix here: https://github.com/TryGhost/extract-zip


AileenCGN:
Uploading a corrupted zip file will look like this now:

image

AileenCGN:
@ErisDS Dependency is updated to https://github.com/TryGhost/extract-zip and ready to go!

PR #110, opened 2018-05-28
Adds @partial-block to helpers list

### LINKED ISSUE #100: Templates must contain valid Handlebars: "The partial @partial-block could not be found"
### Issue Summary

When I use the [`@partial-block`](https://handlebarsjs.com/partials.html#partial-block) feature of handlebars I get this error:

GScan Standalone:
image

```
Checking zip file...

Rule Report:

! Must fix:
- error Templates must contain valid Handlebars

? Consider fixing:
- recommendation <code>package.json</code> property <code>"config.posts_per_page"</code> is recommended. Otherwise, it falls back to 5

✓ 44 Passed Rules

...checks complete.
```

Ghost: 
image

```
Fatal Errors
(Must-fix to activate theme)

Templates must contain valid Handlebars
Oops! You seemed to have used invalid Handlebars syntax. This mostly happens, when you use a helper that is not supported.
See the full list of available helpers here.

Affected files:
partials\header.hbs: The partial @partial-block could not be found
partials\postPartial.hbs: The partial @partial-block could not be found
```

I was previously able to use the [`@partial-block`](https://handlebarsjs.com/partials.html#partial-block) feature of handlebars to pass templates to a partial. Now apparently this does not work in the latest version of Ghost and GScan or even the old version I used to use after I redownloaded (I don't know how I got this to work previously). I did somehow manage to get it on my Ghost Pro blog so it must've worked at some point (cobertos.com)

Ghost Version: `1.22.0` and `1.16.2`
Gscan Version: `1.4.0` and `1.2.1`
Node Version: `8.11.1`

Let me know if this feature is supposed to be supported and if not I'll just close this and I'll switch to a solution that uses two partials.

### Workaround

I came up with a workaround that just involves making two partials that take the same parameters. If you have a `post` partial that uses `@partial-block` like:

```
//In someTemplate.hbs
{{#> postPartial}}
This is {{cool}}
{{/postPartial}}

//In partials/postPartial.hbs
{{> @partial-block}}
```

you would instead do

```
//In someTemplate.hbs
{{> postPartialBegin}}
This is {{cool}}
{{> postPartialEnd}}
```

where `postPartialBegin` contains the content before the `@partial-block` and `postPartialEnd` contains all the content after `@partial-block`

jeffjewiss:
Hi, are you looking to see tests for this as well or just that the helper is added to the list since it should be valid?

Closes #100

### DIFF
--- a/lib/spec.js
+++ b/lib/spec.js
@@ -15,7 +15,7 @@ knownHelpers = [
     'input_email', 'input_password', 'amp_components', 'amp_content', 'amp_ghost_head', 'subscribe_form',
     // Handlebars and express handlebars
     // TODO add a warning that `each` should not be used because Ghost has special behaviour in `foreach`
-    'log', 'if', 'unless', 'with', 'block', 'contentFor', 'each', 'lookup'
+    'log', 'if', 'unless', 'with', 'block', 'contentFor', 'each', 'lookup', '@partial-block'
     // Registering these will break template compile checks
     // 'blockHelperMissing', 'helperMissing',
 ];
--- a/lib/spec.js
+++ b/lib/spec.js
@@ -452,5 +452,5 @@ ruleNext = {
 module.exports = {
     knownHelpers,
     templates,
-    rules,
+    rules
 };
--- a/lib/spec.js
+++ b/lib/spec.js
@@ -4,7 +4,7 @@
  * This file contains details of the theme API spec, in a format that can be used by GScan
  */
 
-var knownHelpers, templates, rules, ruleNext; // eslint-disable-line no-unused-vars
+let knownHelpers, templates, rules, ruleNext; // eslint-disable-line no-unused-vars
 
 knownHelpers = [
     // Ghost
@@ -450,7 +450,7 @@ ruleNext = {
 };
 
 module.exports = {
-    knownHelpers: knownHelpers,
-    templates: templates,
-    rules: rules
+    knownHelpers,
+    templates,
+    rules,
 };
--- a/test/005-template-compile.test.js
+++ b/test/005-template-compile.test.js
@@ -66,4 +66,15 @@ describe('Template compile', function () {
             done();
         }).catch(done);
     });
+
+    it('theme with block partials and unknown helper', function (done) {
+        utils.testCheck(thisCheck, 'theme-with-block-partials').then(function (output) {
+            output.should.be.a.ValidThemeObject();
+
+            output.results.fail.should.be.an.Object().with.keys('GS005-TPL-ERR');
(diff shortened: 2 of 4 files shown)


lib/spec.js
```diff
@@ -15,7 +15,7 @@ knownHelpers = [
     'input_email', 'input_password', 'amp_components', 'amp_content', 'amp_ghost_head', 'subscribe_form',
     // Handlebars and express handlebars
     // TODO add a warning that `each` should not be used because Ghost has special behaviour in `foreach`
-    'log', 'if', 'unless', 'with', 'block', 'contentFor', 'each', 'lookup'
+    'log', 'if', 'unless', 'with', 'block', 'contentFor', 'each', 'lookup', '@partial-block'
```
kirrg001: Could you please add a unit test? Thanks!
Let me know if you need help!

jeffjewiss:
I've added a unit test for using a `@partial-block` helper, but I'm not sure if this is how you'd like to see it tested.

Please let me know when you've had a chance to review and if you'd like the test setup in a different way.

Thanks!

lib/spec.js
```diff
@@ -450,7 +450,7 @@ ruleNext = {
 };
 
 module.exports = {
-    knownHelpers: knownHelpers,
-    templates: templates,
-    rules: rules
+    knownHelpers,
+    templates,
+    rules,
```
kirrg001: 👻

test/fixtures/themes/theme-with-block-partials/index.hbs
```diff
@@ -0,0 +1,3 @@
+{{#> block}}
+{{nope "string"}}
+{{/block}}
```
kirrg001: Some of the test theme files can be removed i guess e.g. `logo.new.hbs` or `subfolder/test.hbs`

lib/spec.js
jeffjewiss: 😬

jeffjewiss:
I fixed the lint error and removed a few of the unnecessary fixture files. Please let me know if you'd like any other changes made. Cheers.

test/005-template-compile.test.js
```diff
@@ -66,4 +66,15 @@ describe('Template compile', function () {
             done();
         }).catch(done);
     });
+
+    it('theme with block partials and unknown helper', function (done) {
```
kirrg001: I am curious why this case tests two things

1. usage of {{> @partial-block}}
2. invalid helper

gargol:
The issue was looked closer into in https://github.com/TryGhost/gscan/pull/227 and merged to master.  Thanks for contribution @jeffjewiss

PR #112, opened 2018-06-17
✨ Read any .hbs file in theme (folder exceptions: assets & partials)

kirrg001:
refs #9601

- custom template feature functionality is untouched
  - we only define custom templates as `custom-{name}.hbs` or `page|post-{name}.hbs` in the root directory of your theme
- ensure we load any custom folder with hbs files into `theme.templates.all`
- ensure we ignore partials & assets folder
- useful feature extension for dynamic routing
  - e.g. `template: podcast/rss`


ErisDS:
I saw this go by, it sounds to me like a breaking change? I thought we supported both root folder & sub folder. Maybe @kevinansfield can review?

kirrg001:
This is not a breaking change in my opinion 🤔

The custom template feature for the PSM makes use of `templates.custom` (a specific response from gscan, which contains `custom-\w+` or `page|post-\w+`). This is untouched. 

And `templates.all` contains any template (except of partials) and is used to look up a template if we want to render with handlebars. This is useful for the `routes.yaml` file - you can access any hbs file from your theme e.g. `template: podcast/rss`.

I agree, the custom template feature in the PSM could show templates living in the folder `custom/`, but this would be a feature extension in my opinion? Let me know what you think.

I would love to merge this for https://github.com/TryGhost/Ghost/pull/9693.

kirrg001:
Summary: Any hbs template we can find in your theme (except of templates which are in the assets folder or in the partials folder) should be added to `templates.all`.

PR #113, opened 2018-06-25
🛠 Updated tooling to use new standards

ErisDS:
cc @kirrg001, @AileenCGN & @kevinansfield I'm moving us towards using these tools everywhere - much much simpler approach, I think.

- Run lint after tests
- Run tests before shipping
- Use yarn & git directly for shipping changes


AileenCGN:
@ErisDS How will this work with private repos? Did you see my comment here: https://github.com/TryGhost/my.ghost.org/pull/209#issuecomment-396157445 ?

PR #114, opened 2018-07-15
Fixed the command line for installing gscan globally with npm

bryan-williams-sls:
There was a typo in the command-line for the npm installation, it was enough to break it for me, the correct option is a lower-case g. Simple fix, might help someone, definitely won't break anything.


kirrg001:
Thanks :)

PR #118, opened 2018-07-25
[HOLD] ✨Allow GScan to scan different versions

### LINKED ISSUE #116: GScan 2.0 should be able to scan a 1.0 theme
GScan 2.0 should be able to:

- scan a theme for Ghost 2.0 (latest)
- scan a theme for Ghost 1.0 with the help of a flag e.g. `gscan --v1`

The idea is that GScan is able to scan a "new" and a "previous" version of Ghost.
If we e.g. ship Ghost 3.0, GScan will scan latest (3.0) and 2.0.

GScan must provide a ruleset for both versions. Currently we are making use of a single `spec.js` file, which holds **most** of the rules. Some of the rules live inside the checkers, which should be moved into the `spec.js` file. 

## Which rules to apply on a 2.0 theme?

I think we have to apply all rules from 2.0 and 1.0 🤔 We just need to ensure that the GScan 2.0 branch doesn't have a rule, which conflicts for 1.0 and 2.0. Theoretical example: We have a deprecation rule in 1.0, which tells the user to use Y instead of X. And in 2.0 we tell the user to remove Y completely, because it's no longer supported. GScan should not show both warnings/errors.

## Tasks
- [x] make it possible to pass a version argument `--v1`, which will tell GScan to scan the theme based on Ghost 1.0 rules
- [x] add a very simple UI to choose latest or v1
- [x] merge into GScan 2.0 branch
- [x] output scanned version (see comment [here](https://github.com/TryGhost/gscan/pull/118#issuecomment-408552474) -> #128 
- [ ] Write logic to remove rules in a newer version (e. g. a prev. known helper should now be removed), see comment [here](https://github.com/TryGhost/gscan/pull/118#discussion_r207781870)
- [x] Central place to adjust the major version (similar to a config value) -> #127 
- [x] Make links to documentation dependent on the version (see comment [here](https://github.com/TryGhost/gscan/pull/118#discussion_r208211785)) -> #127

AileenCGN:
closes #116

- CLI
    - Added `-1, --v1` option to allow cli usage of GScan to check for 1.0 theme compatibility
    - Passed `options` that includes a `checkVersion` property, which is set to `v1` when cli called with `-1`, or `latest` by default.
- Restructure current specs
    - created a new `/specs` directory
    - renamed current `spec.js` to `v1.js`
    - duplicated current `spec.js` to `latest.js` -> will contain only differences later
    - the latest version will inherit the ruleset from `v1`, but overwrite properties that exist in both
- App
    - Offered a select box with the - currently - two accepted versions, whereas the latest is set to default
    - Passed the selected version to gscan
- Checks
    - moveed checks properties (regex, helpername, cssclass) to the rules in spec
    - useed list of rules to specify which rules have to run for each version (concat them, if applicable)
    - require the version of the spec that we need for the checks only

----

Todos:
- [x] Tidy up CSS for new `select` option
- [x] Update text for new version in `index.hbs`
- [x] Refactor specs to use two versions
    - [x] 001 deprecations
    - [x] 002 comment_id
    - [x] 005 template compiler
    - [x] 010 package.json
    - [x] 020 theme structure
    - [x] 030 assets
    - [x] 040 ghost head and foot
- [x] Update v2 rules to only contain differences to v1


app/index.js
```diff
@@ -48,9 +48,14 @@ app.post('/',
             path: req.file.path,
             name: req.file.originalname
         };
+        var options = {
+            checkVersion: req.body.version || 'v2'
```
kirrg001: Alternative naming: "latest"
By that we always have "latest" and "--v1" or "latest" and "--v2" etc.

AileenCGN: I was also thinking of adding a latest and previous property to the config to grab them from there. Does that make sense?

Or would it be better to have only the previous version fixed?

kirrg001:
> wants to merge 11 commits into master from 2.0

Might be better to merge your changes from e.g. "support-v1" into branch GScan "3.0". The goal is to have everything ready in a branch for each repository. master branch stays as is till we ship Ghost 2.0.

app/index.js
kirrg001: > I was also thinking of adding a latest and previous property to the config to grab them from there. Does that make sense?

I wouldn't go that far yet, because if we release Ghost 3.0, we have to put manual work into GScan anyway. I don't imagine, we can just simply reconfigure the config and it will work.

kirrg001: But it would be nice to have it globally defined somewhere, so we don't need to change it in x places.

AileenCGN: > But it would be nice to have it globally defined somewhere, so we don't need to change it in x places.

Yeah, that was my intention behind this. Will look where we can do this.

kirrg001: Cool :)

AileenCGN:
> Might be better to merge your changes from e.g. "support-v1" into branch GScan "3.0". The goal is to have everything ready in a branch for each repository. master branch stays as is till we ship Ghost 2.0.

I don't understand this. I thought I should create a `2.0` branch and get everything there ready? See https://github.com/TryGhost/Team/issues/118#issuecomment-405754890

AileenCGN:
image

I changed the index file now to this ^, but I will make a separate issue for design requirements in GScan 2.0, so we're not blocked on the functionality atm.

AileenCGN:
I opened a new issue for the design and wording: https://github.com/TryGhost/gscan/issues/119

@kirrg001 This is ready to go into the 2.0 branch, if you don't see anything major.

AileenCGN:
Oops. I forgot that it's already in 2.0 now. I will create the other PRs differently.

kirrg001:
@AileenCGN Nice Thanks :) Will review asap.

ErisDS:
> the latest version will inherit the ruleset from v1, but overwrite properties that exist in both

I just want to double check that this means that when testing a theme for 2.0 compatibility, gscan will validate all of the v1 rules AND the new v2 rules?

kirrg001:
> I just want to double check that this means that when testing a theme for 2.0 compatibility, gscan will validate all of the v1 rules AND the new v2 rules?

Yes this is what we have discussed. But the latest version is in theory able to override the previous version rules e.g. rule was a warning in V1, but is a fatal error in V2. We merge the rules together (latest is stronger than previous) and get a full set of rules we have to apply on the themes.

kirrg001:
One thing i have noticed: Maybe it would be cool to show/print the used version (latest or v1) on the report (on the shell and on gscan.ghost.org).

lib/specs/latest.js
```diff
@@ -0,0 +1,28 @@
...
+const previousTemplates = previousSpec.templates;
+const previousRules = previousSpec.rules;
+
+let knownHelpers;
+let templates;
+let rules;
+
+knownHelpers = [];
+
+knownHelpers = _.union(previousKnownHelpers, knownHelpers);
```
kirrg001: Hmm if we want to remove helpers in the latest version, how does that work with `.union`?

Are there any tests who reflect how the merge of the rules works?

AileenCGN: No tests yet for this case.  But you're right, currently with wouldn't work. It probably needs more logic after the `_.union` to mark helpers as `deleted` or so in a newer version and remove them from the array. Or something like this... Will have a 🤔about this.

kirrg001:
> One thing i have noticed: Maybe it would be cool to show/print the used version (latest or v1) on the report (on the shell and on gscan.ghost.org).

Is that something you'll add?

app/tpl/index.hbs
```diff
@@ -3,16 +3,26 @@
         <header class="gh-header">
             <h1>Scan Your Theme</h1>
             <p class="gh-subhead">Upload a zip to check for errors, deprecations and other compatibility issues.</p>
-            <p class="gh-subhead warning"><strong>GScan validates your theme based on the Ghost 1.0 requirements.</strong></p>
-            <a href="https://themes.ghost.org/v1.25.0/docs/changelog#ghost-100" target=_blank class="gh-morelink">
-                See the changelog for Ghost 1.0 themes
+
+            <a href="https://themes.ghost.org/v2.0.0/docs/changelog" target=_blank class="gh-morelink">
```
kirrg001: If we release another minor v2, we would need to change these links again.
I would point them to latest: https://themes.ghost.org/.

AileenCGN:
> Maybe it would be cool to show/print the used version (latest or v1) on the report (on the shell and on gscan.ghost.org)

I will add this. Should be a super easy fix

app/tpl/index.hbs
AileenCGN: Yes, I am aware of that. Still thinking of a better solution for that...

AileenCGN:
Closing this PR as it's in 2.0 already. Still open todos from this task are tracked in #116

PR #127, opened 2018-08-08
Use util to handle versions and update doc links

AileenCGN:
refs #116

- Added a util with a version file, that can take properties and handle `latest` versions and other versions in a more gneric way
- Added functionality that replaces old document links with the new ones, based on the `latest` version number.


lib/specs/latest.js
```diff
@@ -1,25 +1,29 @@
...
 
-rules = {};
-
-rules = _.merge(previousRules, rules);
+// Merge the previous rules into the new rules, but overwrite any specified property,
+// as well as adding any new rule to the spec.
+// Furthermore, replace the usage of the old doc URLs that we're linking to, with the
+// new version.
+rules = _.each(_.merge(previousRules, rules), function replaceDocsUrl(value) {
+    value.details = value.details.replace(prevDocsBaseUrlRegEx, latestDocsBaseUrl);
```
kirrg001: There is a "g" missing the regex. I can still see some 1.25 links when i scan the latest version.

lib/specs/latest.js
```diff
@@ -1,25 +1,29 @@
...
 templates = _.union(previousTemplates, templates);
 
-rules = {};
-
-rules = _.merge(previousRules, rules);
+// Merge the previous rules into the new rules, but overwrite any specified property,
+// as well as adding any new rule to the spec.
+// Furthermore, replace the usage of the old doc URLs that we're linking to, with the
+// new version.
+rules = _.each(_.merge(previousRules, rules), function replaceDocsUrl(value) {
```
kirrg001: When i scan with 1.0, i am not seeing 1.25 links in the report?

AileenCGN: Hmm... I do...

AileenCGN: No. It seems to work fine for me. I just logged the wrong rules. Can you show me an example?

lib/specs/latest.js
AileenCGN: Thanks! Added and pushed. It should be correct now.

lib/specs/latest.js
kirrg001: I pulled again and now it works. No idea 🤷🏼‍♂️

PR #128, opened 2018-08-08
Output the major version that was checked

AileenCGN:
refs #116
needs #127

- uses the versions util to fetch the major version and pass it along our `theme` object
- outputs the tested version for CLI and App usage
- for module usage, GScan returns a `checkedVersion` property with the major version that it was checked for
- updated README to include version usage


AileenCGN:
I will add a bit more to the README, so the usage is clear.

PR #131, opened 2018-08-09
Koenig CSS checks for 2.0

### LINKED ISSUE #115: Ghost 2.0 theme compatibility
Required CSS selectors for 2.0 compatibility:

- `.kg-image-wide`
- `.kg-image-full`

Deprecated CSS selectors:

- `.kg-card-markdown` - this will disappear with Ghost 2.0
  - it's recommended that themes add their own wrapper around the `{{content}}` helper and target that instead if needed
  - we could make it a hard requirement that this selector doesn't exist for a theme to be 2.0 compatible. If the theme has it's own wrapper around `{{content}}` then it would be both 1.0 and 2.0 compatible

---

This needs confirmation but we have talked about adding size selection to image/video embed cards, that would require additional classes:

- `.kg-embed-wide`
- `.kg-embed-full`

I wonder if it's worth implementing those now too so that we don't require themes to add more classes soon after launch. Maybe it also points to having more generic "size" classes so that we don't run into the same issue with other cards later on? /cc @JohnONolan, @peterzimon

AileenCGN:
closes #115 

- Added deprecation for `.kg-card-markdown` CSS class in latest ruleset
- Restructured the test fixtures for deprecation tests to test `v1` and `latest`
- Added tests added  CSS class deprecation

- Added a new checking file that checks the presence of two CSS classes:
    1. `.kg-width-wide`
    2. `.kg-width-full`
- A missing CSS file, or one of the classes missing will cause the theme to be invalid
- The errors are classed as fatal
- Some more consistency for test wordings

---

CLI:
image

Web app:
image


kevinansfield:
I'm wondering if the "replace `.kg-card-markdown`" should be a suggestion rather than a warning?

If a theme is using flexbox (like Casper) it will pretty much be essential to have `.kg-card-markdown` in the CSS for it to be compatible with 1.0 and 2.0 🤔

AileenCGN:
> I'm wondering if the "replace .kg-card-markdown" should be a suggestion rather than a warning?

I can certainly downgrade it to be a `recommendation` rather than a `warning`. Probably worth it to mention the [Slack convo](https://ghost.slack.com/archives/CBTJXD9AB/p1533728874000002?thread_ts=1533727533.000173&cid=CBTJXD9AB) here from yesterday.

kirrg001:
Then i also need to show recommendations in the CLI gscan report 🤔 
Is this really a recommendation?

AileenCGN:
We will leave it as a warning for now, as the CLI does not show recommendations. See Slack convo [here](https://ghost.slack.com/archives/CBTJXD9AB/p1533819310000402)

Wording incl. links to docs will be finalised later, but before the release of 2.0

@kevinansfield feel free to merge this then!

kirrg001:
> Wording incl. links to docs will be finalised later, but before the release of 2.0

Yeah 👍 We'll update docs and docs links next week in the target branches.

@AileenCGN Feel free to self-merge into the 2.0 branch if the PR is blocking you.

We'll do a full QA next week anyway and we won't release GScan 2.0 before the QA has finished. We can update the GScan tarball link in our Ghost 2.0 branch.

@kevinansfield Do you have a theme, which contains common cases and edge cases for the editor css classes? Would be super helpful for testing.

PR #132, opened 2018-08-11
Added dropdown arrow to select box

### LINKED ISSUE #119: GScan 2.0 design and wording
GScan 2.0 will have the ability to choose a version. Therefore we offer a select box now.
Furthermore, the text changed. As a current WIP version, it looks like this:

image

This needs some design and wording refinement (or not, both ways ok)

Note, GScan CSS file lives in ghost.org, but is not referred to directly from GScan, but copied and pasted after each change. See https://github.com/TryGhost/Ghost.org/pull/186

AileenCGN:
closes #119

- Added svg partials for arrow up and down (taken from ghost.org)
- Added toggle on click, but does not really work with native `select` -> could try chosen at some point
- Updated CSS file processed and stored in ghost.org


app/public/js/gscan.js
```diff
@@ -57,6 +57,16 @@
             $('#theme-submit').prop('disabled', !$(this)[0].files.length);
         });
 
+        // Mobile Menu Trigger
+        $('.gh-nav-burger').click(function () {
```
AileenCGN: Added the hamburger code snippet, as it wasn't working. Not that I expect anyone to test themes on a mobile, but you never know 🤷🏼‍♀️

Issue #133, opened 2018-08-11
Detect invalid handlebar expressions in translations


realthargor:
`{{t "page {{page}" page=page}}` does not raise an error, but should.
Valid would be `{{t "page {{page}}" page=page}}`

aileen:
Closing as stale

PR #134, opened 2018-08-11
Detect invalid handlebar expressions in translations (#133)

realthargor:
Reports an invalid handlebar expression in translation as compile error.

I could not find a way to return the correct line inside the template-file.
The handlebar-helpers do not seems to be aware of this.


kirrg001:
Thanks 😎- will review asap, probably after the GScan 2.0 release.

gargol:
This PR refs #https://github.com/TryGhost/gscan/issues/163 which will be solved as a part of a new approach using Handlebars AST parsing started in https://github.com/TryGhost/gscan/pull/218

PR #136, opened 2018-08-12
[WIP] Added ability to remove rules from checks

### LINKED ISSUE #135: GScan should be able to remove checks from previous major versions
The current logic in GScan 2.0 adds new rulesets on top of the old rulesets. There is currently no way to remove a rule, template, or known helper, that we don't want to check for anymore in the latest version.

AileenCGN:
closes #135

- added functionality in the latest spec to exclude helpers, templates, or rules from further checks

Still need to figure out tests...

### DIFF
--- a/lib/checks/002-comment-id.js
+++ b/lib/checks/002-comment-id.js
@@ -15,7 +15,10 @@ checkCommentID = function checkCommentID(theme, options) {
             return rule;
         }
     });
+    console.log('​checkCommentID -> ruleSet', ruleSet);
+
     _.each(ruleSet, function (check, ruleCode) {
+        console.log('​checkCommentID -> ruleCode', ruleCode);
         _.each(theme.files, function (themeFile) {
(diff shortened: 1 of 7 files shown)


ErisDS:
I looked at this to see if I could unblock it and I think it was working already? There's no tests for this check so not sure if that was just confusing, or if I didn't see what was wrong.

Still, on reflection I actually think we should _NOT_ remove these rules.

Original thought:

- people who haven't already changed their `{{id}}` tags obviously didn't need to, so don't remind them again it's annoting

New thought:

- we support the current and previous versions
- people can't upgrade Ghost straight from v0.11 -> v2, but they _can_ upgrade themes that far all in one go
- there are still a lot of people on v0.11 who will want to upgrade, and therefore will use the default gscan checks
- the disqus error is still an error, and the id rule is only a recommendation anyway, so there's no real harm to leaving them.

I think instead we should just change change the wording. I'll PR the wording change, and possibly adjust some of the other wordings whilst I'm there first thing tomorrow.

@AileenCGN you can leave this and move on to other tasks.

PR #153, opened 2018-08-22
🎨 Optimised GScan rule messages

kirrg001:
no issue

- {{author}} is an output helper and it should be replaced with {{authors}} (not with {{primary_author}}, because this is not an output helper, it's an object)
- {{#author}}{{/author}} should be replaced with {{#primary_author}} or {{#foreach authors}}..
- update rule texts to use "is deprecated" instead of "is not valid anymore"

### DIFF
--- a/test/fixtures/themes/001-deprecations/v1/invalid/post.hbs
+++ b/test/fixtures/themes/001-deprecations/v1/invalid/post.hbs
@@ -38,3 +38,5 @@
         {{cover}}
     {{/if}}
 {{/author}}
+
+{{author}}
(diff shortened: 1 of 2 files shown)


test/fixtures/themes/001-deprecations/v1/invalid/post.hbs
```diff
@@ -38,3 +38,5 @@
         {{cover}}
     {{/if}}
 {{/author}}
+
+{{author}}
```
kirrg001: Not sure why the tests didn't complain - @AileenCGN Could you pls double check? Thanks :)

kirrg001:
@AileenCGN Feel free to change whatever you see and merge+release a patch :)

test/fixtures/themes/001-deprecations/v1/invalid/post.hbs
AileenCGN: That's because this fixture is only being tested for v1. Everything good!

PR #158, opened 2018-08-30
Added Koenig CSS class checks for gallery card in 2.1.0

kevinansfield:
no issue
- the gallery card requires CSS for correct display so class name checks were added:
  - `.kg-gallery-container`
  - `.kg-gallery-row`
  - `.kg-gallery-image`


lib/specs/latest.js
```diff
@@ -410,13 +410,39 @@ rules = {
         regex: /\.kg-width-full/g,
         className: '.kg-width-full',
         css: true
-
+    },
+    'GS050-CSS-KGGC': {
+        level: 'error',
+        rule: 'The <code>.kg-gallery-container</code> CSS class is required to use your theme with 2.1',
```
kirrg001: > your theme with 2.1

Should we say something like "your theme with Ghost >= 2.1.0?"

kirrg001: Or not use the version mentioning at all?

PR #172, opened 2018-12-11
Fix temp files not being deleted of unzipped theme in cache 

notanengineercom:
As the issue opened in the Ghost Repo [#10174](https://github.com/TryGhost/Ghost/issues/10174) states, the unzipped files of themes aren't being deleted on failure. This means that when calling gscan, and validating the theme, Ghost just rejects the theme as invalid, and leaves the unzipped theme in the os.tmpdir(). Adding this validation (it's the same that Ghost uses in **/core/server/services/themes/validate.js**), gscan deletes the unzipped dir. If valid theme, Ghost does handle it's deletion without troubles in **/core/server/api/v0.1/themes.js**. 

I added the deletion inside gscan, because I think it makes more sense, gscan does the unzip, so it should delete it's files too. And Ghost cannot get a proper way to access the path to the unzipped theme folder, because this path is in a new object cloned in gscan scope, and Ghost gets this object referenced back only on a successful validation.

### DIFF
--- a/lib/format.js
+++ b/lib/format.js
@@ -152,6 +152,10 @@ format = function format(theme, options = {}) {
         theme.results.error = _.orderBy(theme.results.error, 'fatal', 'desc');
     }
 
+    if (theme.results.error.length || theme.results.hasFatalErrors) {
+        require('./promised-fs').removeDir(theme.path);
+    }
+
     return theme;
 };
 
--- a/lib/format.js
+++ b/lib/format.js
@@ -152,10 +152,6 @@ format = function format(theme, options = {}) {
         theme.results.error = _.orderBy(theme.results.error, 'fatal', 'desc');
     }
 
-    if (theme.results.error.length && options.onlyFatalErrors || theme.results.hasFatalErrors) {
-        require('./promised-fs').removeDir(theme.path);
-    }
-
     return theme;
 };
 
--- a/lib/format.js
+++ b/lib/format.js
@@ -152,7 +152,7 @@ format = function format(theme, options = {}) {
         theme.results.error = _.orderBy(theme.results.error, 'fatal', 'desc');
     }
 
-    if (theme.results.error.length && onlyFatalErrors || theme.results.hasFatalErrors) {
+    if (theme.results.error.length && options.onlyFatalErrors || theme.results.hasFatalErrors) {
         require('./promised-fs').removeDir(theme.path);
     }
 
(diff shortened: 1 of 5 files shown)


lib/format.js
```diff
@@ -152,6 +152,10 @@ format = function format(theme, options = {}) {
         theme.results.error = _.orderBy(theme.results.error, 'fatal', 'desc');
     }
 
+    if (theme.results.error.length && options.onlyFatalErrors || theme.results.hasFatalErrors) {
```
gargol: This seems like not an optimal place to be handling file removal because the function itself should just be doing error formatting. Deleting files seems like a bit too much of a side effect :smiley:  

We also need to be safer with what we delete and allow for optional file removal. Would suggest moving this logic a layer up to https://github.com/TryGhost/gscan/blob/master/lib/index.js#L46, and create a wrapper function with  option similar to `keepExtractedDir` in [here](https://github.com/TryGhost/gscan/blob/master/lib/index.js#L27). This would give us bit more separation of concerns and configurability :+1:

test/general.test.js
```diff
@@ -399,6 +399,14 @@ describe('Checker', function () {
 });
 
 describe('format', function () {
+    beforeEach(function () {
```
gargol: Would be great to have a separate test case that shows that files are being removed when errors are present.

lib/format.js
notanengineercom: > Would suggest moving this logic a layer up to https://github.com/TryGhost/gscan/blob/master/lib/index.js#L46, and create a wrapper function with option similar to keepExtractedDir in here.

The funny thing is that I first thought of doing something similar as your suggestion, but thought that it wasn't the right approach. Now I see that it clearly is, as you say, format.js is just for error formatting. At the end I just searched for the smallest way to fix it, so at the Ghost Core there wouldn't be that much to change. 
So thanks for the feedback, I had some busy days, but tomorrow I'll be able to work on it 👍 
The only question I have for now, is that the optional file removal should be enabled by default, right?

test/general.test.js
notanengineercom: Yes, that would be better, easier to maintain!

notanengineercom:
@gargol, this is a really easy fix, and I have the fix for Ghost ready, could be still added to the open PR there.
It's just a line of code on **core/server/services/themes/validate.js**

gargol:
The improvement on removing temporary files have been done and merged on Ghost's side: https://github.com/TryGhost/Ghost/pull/10274. 

After some more investigation, the only way `gscan` could be leaving behind temporary files is when it's being **explicitly** told to do so with `keepExtractedDir` parameter, see this [comment](https://github.com/TryGhost/gscan/pull/172#discussion_r242142210).

No further changes are needed here :smiley:

PR #176, opened 2019-03-05
Add Dockerfile for building an image on node 10

pysysops:
Hi,

At some point in the near future I'd like to run gscan and some of our other services on docker and maybe even Kubernetes.

This PR adds a Dockerfile for building gscan as a container image.

To try this out, checkout this branch.

### Build the image
```
docker build -t gscan .
```

### Run the image locally
```
docker run -it -p 2369:2369 gscan:latest
```

Overriding the port that gscan runs on:
```
docker run -it -e port=8080 -p 8080:8080 gscan:latest
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,10 @@
+node_modules/
+*.log
+.git*
+.npmignore
+.travis.yml
+circle.yml
+.dockerignore
+Dockerfile
+nodemon.json
+shipitfile.js
--- a/.npmignore
+++ b/.npmignore
@@ -4,3 +4,5 @@ test/**
 .travis.yml
 Gruntfile.js
 node_modules/**
+.dockerignore
+Dockerfile
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
 FROM node:10-alpine
 
+ENV NODE_ENV development
+
 # Set port to run application on
 ENV port 2369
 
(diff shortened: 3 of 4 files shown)


pysysops:
I feel like that will create a larger image.
```
RUN yarn install
```
would create a layer with all deps installed
```
RUN chown -R $user:$user /usr/src/app \
    && yarn install && yarn test \
    && yarn install --production && yarn cache clean
```
Would create a single layer after adding deps, testing, removing dev dependencies and clearing cache.

Your example may optimise for downloading of dependencies but results in a larger image as the layer has the yarn cache and dev dependencies within it.

My Dockerfile is optimised for image size:
```
docker images gscan*
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
gscanegg            latest              c083111ebc45        20 seconds ago      196MB
gscan               latest              74589497122b        27 minutes ago      97MB
```

allouis:
Oh cool - I didn't realise those weren't flattened out!

I've always optimized for speed because I used to dev inside of docker! For this purpose tho less layers is deffo better - thanks for sharing

pysysops:
What _could_ be done is a multi-stage build copying over files from the builder container _but_  I feel that would be an over-optimisation for this simple service.

Every directive in a Dockerfile creates a layer. Even if you remove files further down the Dockerfile, they still exist in the earlier layer.

It's not uncommon to have multiple Dockerfiles in a repo, one for development and one for deployment.

pysysops:
Mounting config into container:
```
docker run -it -v `pwd`/config.development.json:/usr/src/app/config.development.json:ro -p 2369:2369 gscan:latest
```

gargol:
Hey @pysysops should we merge these changes? Is there some bigger plan/issue that it could be linked to?

gargol:
ping @pysysops :wink:

gargol:
Dockerization of the project was put on hold for now. Might be a good starting point if the idea comes back int the future :+1:

PR #180, opened 2019-04-01
{{blog}} and {{lang}} helpers deprecation

### LINKED ISSUE #174: Add deprecation notice for {{@blog}} and {{lang}}
We're renaming `{{@blog}}` to `{{@site}}` prior to the Content API launch. 

We're also going to remove the superfluous `{{lang}}` helper, as this is a duplicate of `{{@site.lang}}`, which is also clearer. 

Both of these things will continue to work for a while, but we should add deprecation notices ahead of time, and prepare to remove them entirely in Ghost 3.0.

gargol:
closes #174

@ErisDS would love a review of error message wording for 2 new rules GS001-DEPR-BLG and GS001-DEPR-LNG. Thanks!

### DIFF
--- a/lib/specs/latest.js
+++ b/lib/specs/latest.js
@@ -374,6 +374,15 @@ rules = {
         regex: /{{\s*?(?:error\.)?(statusCode)\s*?}}/g,
         helper: '{{error.statusCode}}'
     },
+    'GS001-DEPR-BLG': {
+        level: 'error',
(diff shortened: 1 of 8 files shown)


ErisDS:
@gargol I updated the wording, and also changed the rule names, as I don't think there's any need to remove the vowels? This would need updating elsewhere.

PR #201, opened 2019-04-05
🛠 Ported cli to use pretty-cli

### LINKED ISSUE #177: Migrate to pretty-cli
All the cli tooling that Ghost relies on has to have a consistent interface - [pretty-cli](https://www.npmjs.com/package/@tryghost/pretty-cli). We are aiming to migrate `gscan` as well.

gargol:
closes #177

@ErisDS one thing that bothers me with new output for usage:
```
Usage:
  gscan <themePath> [options]

Arguments:
  <themePath>  Theme folder or .zip file path                                    [required] [string]

Global Options:
  -h, --help     Show help                                                [commands: help] [boolean]
  -v, --version  Show version number                                   [commands: version] [boolean]

Options:
  -p, --pre  Run a pre-check only                                                          [boolean]
  -z, --zip  Theme path points to a zip file                                               [boolean]
  -1, --v1   Check theme for Ghost 1.0 compatibility, instead of 2.0                       [boolean]
```
is that `Options:` part, that the user would be most interested in, is listed in the very end of the output. Haven't found a good way to override that in `sywac` (other than listing everything in `positional(...)` configuration), maybe you know a way?

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -1,26 +1,67 @@
 #!/usr/bin/env node
-
-const pkgJson = require('../package.json');
-const program = require('commander');
+/* eslint-disable no-console */
+const prettyCLI = require('@tryghost/pretty-cli');
 const _ = require('lodash');
 const chalk = require('chalk');
 const gscan = require('../lib');
 
 const options = {};
-let themePath = '';
 let levels;
 
-program
-    .version(pkgJson.version)
-    .usage('[options] <themePath>')
-    .arguments('cmd <themePath>')
-    .option('-p, --pre', 'Run a pre-check only')
-    .option('-z, --zip', 'Theme path points to a zip file')
-    .option('-1, --v1', 'Check theme for Ghost 1.0 compatibility, instead of 2.0')
-    .action(function (theme) {
-        themePath = theme;
+prettyCLI
+    .configure({
+        name: 'gscan'
     })
-    .parse(process.argv);
+    .groupOrder([
+        'Sources:',
+        'Utilities:',
+        'Commands:',
+        'Arguments:',
+        'Required Options:',
+        'Options:',
+        'Global Options:'
+    ])
+    .positional('<themePath>', {
+        paramsDesc: 'Theme folder or .zip file path',
+        mustExist: true
+    })
+    .boolean('-p, --pre', {
+        desc: 'Run a pre-check only'
+    })
+    .boolean('-z, --zip', {
+        desc: 'Theme path points to a zip file'
+    })
+    .boolean('-1, --v1', {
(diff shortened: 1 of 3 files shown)


ErisDS:
@naz it's an option called `groupOrder`:

E.g. https://github.com/TryGhost/migrate/blob/master/packages/migrate/bin/cli.js#L11

gargol:
@ErisDS that's perfect! thx :heart:

gargol:
Think it's ready to be merged :ship:

ErisDS:
FYI all the loglines like this:
https://github.com/TryGhost/gscan/blob/master/bin/cli.js#L74

Should be able to use pretty-cli's `ui.log.level()` methods:

https://github.com/TryGhost/Ghost-Utils/blob/master/packages/pretty-cli/lib/ui.js

E.g. https://github.com/TryGhost/migrate/blob/master/packages/migrate/commands/medium.js#L52

I've updated the pretty-cli README.md to mention this:

https://github.com/TryGhost/Ghost-Utils/blob/master/packages/pretty-cli/README.md

ErisDS:
P.S. I think that pretty-cli should also expose chalk, that way it doesn't need to be added as an extra dependency

gargol:
@ErisDS looked into changing `console.log` to `ui.log.{level}()` today and one thing that I'm very not sure about is the change in formatting log level introduces. For example, previously we would have following output:
```
> node ./bin/cli.js test/fixtures/themes/bad-example.zip --zip
Checking zip file...

Rule Report for v2.x:
```
and with `ui.log.info` this changes to :
```
>node ./bin/cli.js test/fixtures/themes/bad-example.zip --zip
info Checking zip file...
info Rule Report for v2.x:
```
The level prefixes become extra hard on the ui when the report is displayed and becomes:
```
! Must fix:
info - error <code>package.json</code> property <code>"author.email"</code> is required
info     Files: package.json
info - error <code>package.json</code> property <code>"author.email"</code> must be valid
info     Files: package.json
info - error The <code>.kg-width-wide</code> CSS class is required to use your theme with 2.0
info     Files: styles
info - error The <code>.kg-width-full</code> CSS class is required to use your theme with 2.0
info     Files: styles
info - error The <code>.kg-gallery-container</code> CSS class is required to use your theme with 2.0
info     Files: styles
....
```

For now substituted `console.log` usages with just `ui.log` ([an alias](https://github.com/TryGhost/Ghost-Utils/blob/30cee77/packages/pretty-cli/lib/ui.js#L2) to console.log). 

> P.S. I think that pretty-cli should also expose chalk, that way it doesn't need to be added as an extra dependency

I agree with this. Additionally, we should probably define a set of standard formattings e.g.: `chalk.yellow.bold.underline` for `info` so that we have a standard set to use across pretty-cli clients :+1:

PR #209, opened 2019-04-11
Updated results page styles

peterzimon:
refs https://github.com/TryGhost/Team/issues/206

- changed the overall style of the result page to improve readability

### DIFF
--- a/app/tpl/result.hbs
+++ b/app/tpl/result.hbs
@@ -1,28 +1,26 @@
 <section class="gh-section report">
     <div class="inner">
         <header>
-            <h2 class="report-heading">
+            <h1 class="report-heading">
                 <div class="result-summary">
-                    <span class="score {{results.score.level}}">{{results.score.value}} / 100</span>
-                    {{#if name}}
-                        <span class="title">{{name}} {{version}}</span>
-                    {{/if}}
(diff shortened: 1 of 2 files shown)


app/tpl/result.hbs
```diff
@@ -1,28 +1,26 @@
...
+                <span class="score {{results.score.level}}">{{results.score.value}} / 100</span>
+            </div>
         </header>
     </div>
 
     <div class="inner report-body">
         <div class="results report-section">
-            {{!-- TODO: make this pretty --}}
-            <div>
-                <p>Result for version {{checkedVersion}}:</p>
```
AileenCGN: @peterzimon Do we not want to show which version the theme was checked for?

app/tpl/result.hbs
```diff
@@ -8,7 +8,7 @@
                 <a href="/" class="gh-btn gh-btn-blue"><span>Start Over</span></a>
             </h1>
             <div class="health-meter {{results.score.level}}">
-                <h4>Theme compatibility score</h4>
+                <h4>Theme compatibility for Ghost version {{checkedVersion}}</h4>
```
peterzimon: @AileenCGN do you think this makes more sense?

AileenCGN: I think so. Because the user has to select the version that they want to test it again on the starting screen. IMO this would e. g. flag up if they selected a wrong version unintentionally.

PR #216, opened 2019-04-12
Verbose output option in CLI

### LINKED ISSUE #214: Add error details to CLI output
### Problem
There is no way to see details of the error in CLI output

### Design
image_preview

Changes:
- [x] add --verbose option to output check details
- [x] add `<a href=link>click here</a>` formatting transform

gargol:
closes #214

The output of the CLI check with `--verbose` parameter:
Screenshot from 2019-04-12 13-05-44

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -82,7 +82,7 @@ function outputResult(result) {
     ui.log(levels[result.level](`- ${_.capitalize(result.level)}:`), result.rule);
 
     if (options.verbose) {
-        ui.log(`${chalk.bold('\nDetails: ')} result.details`);
+        ui.log(`${chalk.bold('\nDetails:')} ${result.details}`);
     }
 
     if (result.failures && result.failures.length) {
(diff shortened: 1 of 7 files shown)


peterzimon:
@gargol I think this is good, just two minor things:
1. could we add an extra line after each "Files:..." row (only when `--verbose` is on)? that would make it easier to understand which detail belongs to which error/warning
2. there seem to be an extra `.onelineTrim` at the end of each detail info. looks like it's a bug?

gargol:
@peterzimon updated the spacing :) the other issue with `.onelineTrim` was a temporary bug which was fixed right away but forgot to update the screenshot. Here's how the output looks like now:

Screenshot from 2019-04-16 12-35-33

Screenshot from 2019-04-16 12-32-29

Let me know if it looks acceptable and I'll merge the change :+1:

peterzimon:
Looks 👍 thank you!

PR #218, opened 2019-04-12
Initial AST based linter spike

kevinansfield:
This is the beginning of a new method of checking templates by walking the Handlebars AST rather than performing regexes.

There's still a fair bit to do to integrate with the rest of gscan but it looks like a promising start. There are two example rules implemented as well as a `test.js` file that looks like this:

```js
const Linter = require('./');

const linter = new Linter();
const template = `
    {{test}}
    {{#if img_url feature_image}}
        {{feature_image}}
    {{/if}}
`;

const messages = linter.verify({
    source: template,
    moduleId: 'test.hbs'
});

console.log(JSON.stringify(messages));
```

When run, this is the logged output:

```
❯ node lib/ast-linter/test.js | jq
[
  {
    "moduleId": "test.hbs",
    "rule": "no-img-url-in-conditionals",
    "message": "The {{img_url}} helper should not be used as a parameter to {{#if}} or {{#unless}}",
    "line": 3,
    "column": 4,
    "source": "{{#if img_url feature_image}}        {{feature_image}}    {{/if}}",
    "fix": "Remove the 'img_url' so your conditional looks like {{#if feature_image}}"
  },
  {
    "moduleId": "test.hbs",
    "rule": "no-multi-param-conditionals",
    "message": "Multiple params are not supported in an {{if}} or {{unless}} statement.",
    "line": 3,
    "column": 4,
    "source": "{{#if img_url feature_image}}        {{feature_image}}    {{/if}}"
  }
]
```


gargol:
@kevinansfield happy to merge this PR as is and have created a placeholder issue https://github.com/TryGhost/gscan/issues/265 to track progress with other improvements. Plan to write any new rules that will come into v3 set using this new method if possible. Lmk what you think ;)

PR #241, opened 2019-07-25
Initial AST based linter spike (with context tracking)

kevinansfield:
Follow-on from https://github.com/TryGhost/gscan/pull/218.

Opened as a separate PR because this is in a partially finished broken state whereas #218 _is_ workable for syntax-only rules.


naz:
The code is not finished but it's passing and will be easier to colab on in the main branch. Merging as is with anticipation to improve.

Issue #144, opened 2018-08-13
Ghost 3.0 gscan changes


ErisDS:
### Running v3 rules (`--canary` flag feature)

This flag is needed to allow us testing gscan v3 rules before Ghost v3 is released.

The file with `canary` definitions can be aliased to `--v3` flag, so as we near new version release, we could make it possible to do `gscan --canary` or `gscan --v3`, so that theme developers can have their themes ready against any known changes way ahead of the release.

`canary` spec will exist perpetually (similar to `/canary/` API in Ghost) out of which stable v3/v4 etc. specs will be branched once they reach stability.

Todos:
- [x] add `--canary` option in CLI
- [x] add `Canary (next)` option in Web interface
- [x] make it possible to pass `vX` (`v2`, `v3`) option for `check` function in `checkVersion` options parameter
- [x] add `canary` rule -  warning to check for missing `engines` flag in `package.json` so the API is "falling back" to the default version (https://github.com/TryGhost/gscan/pull/255). 
- [x] update wording in `canary` for to `users` -> `authors` deprecation rule (GS001-DEPR-USER-GET') as this alias has been removed completely (this shoudl be clarified in the update) (https://github.com/TryGhost/gscan/pull/256)
- [x] detect "ghost-api === v1" used in pakcage.json (https://github.com/TryGhost/gscan/pull/258)
- [x] override `GS001-DEPR-ESC` and deprecate use of `{{(error.)?code}}` helper instead of `{{statusCode}}` (https://github.com/TryGhost/gscan/pull/259)
- [x] add "bookmark card classes" rules
- [x] reword rules using "Ghost 2.0" wording
  - [x] check `GS002-DISQUS-ID` and `GS002-ID-HELPER` using "Ghost v1.0" wording

Post v3 release:
- [ ] prepare v3 (out of `canary` rules) set of rules becoming a default (latest) version on a `v3` branch (will release 3.0.0 out of this branch)
  - [x] change Ghsot core from using `canary` flag to default once switched to `gscan 3.0.0` (https://github.com/TryGhost/gscan/pull/257)
  - [ ] merge ^ into master
- [ ] Update details for {{subscribe_form}} deprecation

NOTE: the initial issue was more like a note and can be checked for reference in history.

gargol:
First rules to be added as fatal for v3.0 can be based on https://github.com/TryGhost/gscan/issues/181 and https://github.com/TryGhost/gscan/issues/174.

PR #250, opened 2019-09-04
✨ Version specific checks --v2 and --v3

gargol:
refs #144 

This change adds the ability to run checks against specific version using a flag similar to existing `-1/`--v1`. Following new options will now be available:
- `--v2` (aliased to `--latest`)
- `--v3` (aliased to `--canary`)

`latest` & `canary` are now acting as aliases to versions instead of "defining" them

Worth noting, change does not break current behavior from CLI/lib's API perspective in any way. It only adds 2 "new" parameters in both clients.

@peterzimon there are some words as well as parameter names that need checking in README.md and in bin/cli.js files.

@kevinansfield just looking for input from technical power-user of the CLI perspective (if options make sense) and if new file structure for rules is good)


bin/cli.js
```diff
@@ -35,10 +35,16 @@ prettyCLI
         desc: 'Theme path points to a zip file'
     })
     .boolean('-1, --v1', {
-        desc: 'Check theme for Ghost 1.0 compatibility, instead of 2.0'
+        desc: 'Check theme for Ghost 1.0 compatibility'
+    })
+    .boolean('-2, --v2', {
```
gargol: Another approach to specifying the version could be something like: `gscan --version 2` or `gscan --v 2` or some other name instead of `version` as it's reserved for current script's version check. Don't like this approach as it's more typing and breaks the existing convention.

peterzimon: I agree. the wording looks good to me

bin/cli.js
```diff
@@ -35,10 +35,16 @@ prettyCLI
...
+        desc: 'Check theme for Ghost 2.0 compatibility'
+    })
+    .boolean('-3, --v3', {
+        desc: 'Check theme for Ghost 3.0 compatibility'
     })
     .boolean('-c, --canary', {
-        desc: 'Check theme for upcoming version compatibility, instead of 2.0'
+        desc: 'Check theme for upcoming Ghost 3.0 compatibility, instead of 2.0'
     })
```
peterzimon: to be more generic I would maybe go with:
"Check theme for upcoming Ghost version compatibility"

README.md
```diff
@@ -36,15 +36,27 @@ To run a local zip file through the checks:
...
 `gscan /path/to/theme/directory --v1`
 
+or
+
+`gscan /path/to/theme/directory --v3`
+
+Another accepted parameter is `--canary` which checks for upcoming (3.0) version:
+
+`gscan /path/to/theme/directory --canary`
+
```
peterzimon: My suggestion for the section:
----

### 2. CLI usage
Install using yarn / npm:
`yarn global add gscan` /  `npm install -g gscan`

To run a local directory through the checks:
`gscan /path/to/theme/directory`

To run a local zip file through the checks:
`gscan /path/to/theme.zip -z`

By default, GScan scans themes for the latest Ghost version compatibility. You can also specify a Ghost version by using the following parameters (for Ghost 1.0, 2.0 and 3.0):

`-1` or `--v1`
`-2` or `--v2`
`-3` or `--v3`

Use the `--canary` parameter to check for the upcoming Ghost version.

Examples:

`gscan /path/to/theme.zip -z1` -- scan a theme in a zip file for Ghost 1.0 compatibility
`gscan /path/to/theme/directory --v2` -- scan a theme in a directory for Ghost 2.0 compatibility
`gscan /path/to/theme/directory --canary` -- scan a theme for the upcoming version of Ghost

peterzimon: @JohnONolan could you please doublecheck if the above makes sense? thank you

bin/cli.js
gargol: Agree and prefer the generic wording :+1: `Ghost 1.0` etc. were used in other commands so didn't know what's the best way here :wink:

gargol:
@peterzimon have pushed the edits to readme file. lmk if I got the formatting right (easiest to check here https://github.com/TryGhost/gscan/blob/24396ceebb74e5d73da90ab8c76db20cda70aac1/README.md#2-cli-usaged). think we are ready to merge this :)

README.md
daviddarnes: @peterzimon @gargol does this read better do you think?

### 2. CLI usage

Install using `yarn` or `npm`:
```
yarn global add gscan
```

To check a theme locally:
```
gscan /path/to/theme/directory
```

To check a theme locally within a zip file:
```
gscan /path/to/theme.zip -z
```

By default GScan checks themes against the latest version of Ghost. Flags can be used to check themes against a specific version of Ghost:
- `-1` or `--v1`
- `-2` or `--v2`
- `-3` or `--v3`

Use the `--canary` flag to check a theme against an upcoming version of Ghost.
```
gscan /path/to/theme/directory --canary
```

gargol: @daviddarnes the code sections - "```", read a bit better. Not a fan of those bullet lists tbh :) soft line break seems less distracting when there's not much content in the list to show.

daviddarnes: @gargol I'm cool with that 😄

gargol: the other bit that doesn't hold up with code sections - "```", is the npm/yarn installation section... think for consistency sake and to keep the "examples" section will stick to @peterzimon 's style. welcome to PR again if it feels off.

daviddarnes: Fair enough, I'm not precious about the formatting. I only put yarn as it's a common pattern we do in the docs, we mention npm briefly but we use yarn for the majority of the examples

Issue #208, opened 2019-04-11
Gscan crashes when rule is missing `level` attribute


gargol:
### What's wrong
When a rule that is not using `level` attirbute (like [this one](https://github.com/TryGhost/gscan/blob/fb9e3d7/lib/specs/latest.js#L540)) is encountered gscan crashes.

### Reproduction
To reproduce add `{{primary_author.image}}` any of the templates the theme.

### Implications
In Ghost this causes an error during startup:
```
InternalServerError: Unable to activate the theme "casper".

....

TypeError: Cannot read property 'partials' of undefined
    at new ActiveTheme (/home/gargol/workspace/ghost/Ghost/core/server/services/themes/active.js:39:39)
    at Object.set (/home/gargol/workspace/ghost/Ghost/core/server/services/themes/active.js:124:30)
    at Object.activate (/home/gargol/workspace/ghost/Ghost/core/server/services/themes/index.js:101:20)
    at validationFailure (/home/gargol/workspace/ghost/Ghost/core/server/services/themes/index.js:58:30)
```

### Possible fix
Add `level` attributes to all rules.

PR #267, opened 2019-10-25
Fixed crash when rule level is missing

achudinovskyh:
Added levels to all rules. #208


gargol:
Yessss! Thanks @achudinovskyh! These look just right :tada:

PR #296, opened 2020-03-18
✨ added check for invalid theme translations

### LINKED ISSUE #295: Run a JSON parser on locale files
ref https://github.com/TryGhost/Ghost/issues/11464

Warn people when they upload broken locale files, so that it isn't hidden.

vikaspotluri123:
closes #295

I think the copy needs a look, and I've only added the rule to latest for now


ErisDS:
This looks good! However, we can't introduced fatal errors outside of a major version bump as it would break sites that already have invalid translation files.

A normal error will show up in the right places for now, and then we can upgrade this to fatal in the next major bump.

PR #371, opened 2021-02-05
💡 Split v3 and canary to prepare for the upcoming ghost release

tpatel:
issue https://github.com/TryGhost/Team/issues/221


lib/checker.js
```diff
@@ -18,13 +18,11 @@ const checks = requireDir('./checks');
...
 
-    const passedVersion = _.get(options, 'checkVersion', 'v2');
+    const passedVersion = _.get(options, 'checkVersion', 'v3');
     let version = passedVersion;
 
     // TOOD: get rid of 'latest' concept because its unnecessary invention
-    if (passedVersion === 'latest') {
-        version = 'canary';
-    } else if (passedVersion === 'v3') {
+    if (passedVersion === 'latest' || passedVersion === 'v4') {
```
naz, line 25: Maybe it's time to get rid of the `latest` as suggested above because we are going to bump the major version of `gscan` anyway?

naz:
@tpatel just to make sure I'm understanding the change correctly. What has been done here was:
1. aliased `canary` rules to `v4`
2.  copied over specs & tests from `canary` to `v3`
3. and made `v3` a default version to check the theme against

Is that an accurate summary?

tpatel:
@naz You're very close, here is what I've done:

1. Copied the `canary` code and test to `v3`
2. Aliased `v4` to point on `canary`
3. Modified the package-json logic to include the `v3` tests

I think we should update the tests to split the version by files (easier version update) and the package-json logic to avoid having to modify code when adding/removing a version.

I'm ok to remove "latest", although it's better if it was deprecated in the current gscan version, what do you think?

daniellockyer (approved):
This LGTM - I'm indifferent on the `latest` removal but we can unblock ourselves with this 🙂

tpatel:
I've replaced `latest` by `canary` in https://github.com/TryGhost/gscan/commit/43c5b611f8b5b6a3a5377827b9f72a6105396f49 ✅

PR #375, opened 2021-03-12
💡 Updated theme translation validity check to be fatal when failing

tpatel:
no issue

### DIFF
--- a/test/070-theme-translations.test.js
+++ b/test/070-theme-translations.test.js
@@ -2,6 +2,9 @@ const should = require('should'); // eslint-disable-line no-unused-vars
 const utils = require('./utils');
 const thisCheck = require('../lib/checks/070-theme-translations');
 
+const checker = require('../lib/checker');
+const format = require('../lib/format');
+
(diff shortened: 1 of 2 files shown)


test/070-theme-translations.test.js
```diff
@@ -2,6 +2,10 @@ const should = require('should'); // eslint-disable-line no-unused-vars
 const utils = require('./utils');
 const thisCheck = require('../lib/checks/070-theme-translations');
 
+const themePath = require('./utils').themePath;
```
allouis, line 5: We already require this module in a few lines above this - not a big deal but probs cleaner to just use `utils.themePath`

allouis (approved):
IMO These tests are an improvement as they use the `lib/checker` which is exposed to the public API

rishabhgrg (approved):

test/070-theme-translations.test.js
tpatel: good catch, thanks!

PR #378, opened 2021-04-15
Fix spacing mistakes

stromfeldt:
I recently passed my theme through GScan and noticed that there were a few spacing errors as well as an extra full stop.

GScan

Looking through the relevant file and other related files I noticed a few more typos in the same vein. Here's a bit of an attempt to rectify those.

### DIFF
--- a/lib/specs/canary.js
+++ b/lib/specs/canary.js
@@ -40,7 +40,7 @@ let rules = {
         level: 'warning',
         rule: 'The <code>{{@labs.members}}</code> helper should not be used.',
         details: oneLineTrim`Please remove <code>{{@labs.members}}</code> from the theme.<br>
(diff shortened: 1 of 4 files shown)


stromfeldt:
Forgot to mention: the line breaks that I amended were due to the fact that there seemed to be two design methodologies utilized, which in effect weren't consistent and which I tried to rectify.

ErisDS:
Hey @stromfeldt, thank you so much for taking the time to raise this. I'm so sorry we got really behind on triaging last year 😬 

Would you be up for rebasing your changes? Then we can get this merged.

stromfeldt:
Hey @ErisDS, alright, I finished the rebase, which I hope I did correctly as I'm not exactly all too familiar with all this. 😅

PR #389, opened 2021-06-10
Added example spike tracking missing partials in code flows

naz:
:warning: WIP! :warning: 

refs https://github.com/TryGhost/Team/issues/774

- This is unfinished spike experimenting with AST parser. It is not fully working and the concept could be completely flowed as we have to privide additional state for the AST Linter

### DIFF
--- a/lib/ast-linter/helpers/index.js
+++ b/lib/ast-linter/helpers/index.js
@@ -10,7 +10,16 @@ function getNodeName(node) {
 
     case 'PathExpression':
     default:
-        return node.parts[0];
+        if (node.parts) {
+            return node.parts[0];
+        }
+        if (node.name.parts) {
+            return node.name.parts[0];
+        }
+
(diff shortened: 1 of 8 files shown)


lib/ast-linter/rules/link-no-unknown-partials.js
```diff
@@ -0,0 +1,23 @@
...
+                    column: node.loc && node.loc.start.column,
+                    source: this.sourceForNode(node)
+                });
+            }
+        }
+    }
+
+    visitor() {
+        return {
+            PartialStatement: this._checkForUnknownPartials.bind(this)
```
tpatel, line 20: I would add `PartialBlockStatement` here to collect all partials.

The reason is in the Handlebars spec : a partial can be called as a block to provide a fallback in case the partial isn’t found (https://handlebarsjs.com/guide/partials.html#partial-blocks).

PR #437, opened 2022-05-24
Updated check for deprecated @price usage

rishabhgrg:
The `@price` data helper was only checked when used as a standalone helper - `{{@price.monthly}}`. This wasn't taking into account other usages of `@price` inside other helpers like - `{{price @price.monthly}} or `{{price currency=@price.currency}}` which are also invalid usages.

- expands `@price` check to use regex and flag any usage of `@price` inside the helper brackets


ErisDS:
Superceded by https://github.com/TryGhost/gscan/pull/438

PR #458, opened 2022-11-21
Update error for missing koenig styles

royalfig:
image

When recently helping a customer update to v5, Jon got hit with this error message, which he found confusing for him and the customer. In particular, the bit about `affected files: styles`. The difficulty here is that the error message is not about a specific file  — but about missing files.

This commit updates the error message to be more explicit:
`Missing styles or stylesheets`

### DIFF
--- a/lib/checks/050-koenig-css-classes.js
+++ b/lib/checks/050-koenig-css-classes.js
@@ -63,7 +63,7 @@ const checkKoenigCssClasses = function checkKoenigCssClasses(theme, options) {
             theme.results.fail[ruleCode] = {};
             theme.results.fail[ruleCode].failures = [
                 {
-                    ref: 'Missing styles or stylesheets'
+                    ref: 'Missing stylesheets with required class styles'
                 }
             ];
         }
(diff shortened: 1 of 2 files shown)


lib/checks/050-koenig-css-classes.js
```diff
@@ -63,7 +63,7 @@ const checkKoenigCssClasses = function checkKoenigCssClasses(theme, options) {
             theme.results.fail[ruleCode] = {};
             theme.results.fail[ruleCode].failures = [
                 {
-                    ref: 'styles'
+                    ref: 'Missing styles or stylesheets'
```
naz, line 66: This message appears under "Affected files" section, so would read like:
```
Affected files:
- Missing styles or stylesheets
```

I think "Missing stylesheets with required class styles" would be a more suitable reference here.

royalfig: Agreed! Updated reference text

naz:
@royalfig did you check how these changes appear in Ghost Admin?

PR #466, opened 2023-02-07
update error msg when passing zip without flag

royalfig:
Ref #465 

When passing a zip file to gscan without including the `-z` flag, gscan currently quits without logging the error, which is confusing to users.

This change logs the error to the console, and, if the error is because the path passed is not a dir, then provides a hint about the `z` flag.

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -113,8 +113,10 @@ prettyCLI
             gscan.check(argv.themePath, cliOptions)
                 .then(theme => outputResults(theme, cliOptions))
                 .catch((err) => {
-                    ui.log(`\n${chalk.red(err.message)}`);
-                    err.code === 'ENOTDIR' && ui.log('Did you mean to add the -z flag to read a zip file?');
+                    ui.log(err.message);
+                    if (err.code === 'ENOTDIR') {
+                        ui.log('Did you mean to add the -z flag to read a zip file?');
+                    }
                 });
         }
     });
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -112,11 +112,9 @@ prettyCLI
         } else {
             gscan.check(argv.themePath, cliOptions)
                 .then(theme => outputResults(theme, cliOptions))
-                .catch(function ENOTDIRPredicate(err) {
-                    return err.code === 'ENOTDIR';
-                }, function (err) {
-                    ui.log(err.message);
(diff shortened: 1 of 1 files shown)


bin/cli.js
```diff
@@ -112,11 +112,9 @@ prettyCLI
...
             gscan.check(argv.themePath, cliOptions)
                 .then(theme => outputResults(theme, cliOptions))
-                .catch(function ENOTDIRPredicate(err) {
-                    return err.code === 'ENOTDIR';
-                }, function (err) {
-                    ui.log(err.message);
-                    ui.log('Did you mean to add the -z flag to read a zip file?');
+                .catch((err) => {
+                    ui.log(`\n${chalk.red(err.message)}`);
+                    err.code === 'ENOTDIR' && ui.log('Did you mean to add the -z flag to read a zip file?');
```
naz, line 117: Would be preferable to avoid the `&&` shourtcut notation, it is pretty hard to read unless one knows how JS works. The simpler the better - regular `if` works just fine in this situation 😄

bin/cli.js
```diff
@@ -112,11 +112,9 @@ prettyCLI
         } else {
             gscan.check(argv.themePath, cliOptions)
                 .then(theme => outputResults(theme, cliOptions))
-                .catch(function ENOTDIRPredicate(err) {
-                    return err.code === 'ENOTDIR';
-                }, function (err) {
-                    ui.log(err.message);
-                    ui.log('Did you mean to add the -z flag to read a zip file?');
+                .catch((err) => {
+                    ui.log(`\n${chalk.red(err.message)}`);
```
naz, line 116: Hm. Would be good to check the style consistency here imo. We don't seem to log errors this way anywhere else in the CLI. The catch block above for example just uses 'ui.log'. Also, there's the `levels` configuration up above that configures the "error" [log to be red](https://github.com/TryGhost/gscan/pull/466/files#diff-ca900686fca4680d4692bf587dc5da1322879c344688c187b5511cda27902c18R14). 

Do we only log "theme" errors as red elsewhere?

naz (approved):
For the record the `ENOTDIR` handling comes [from pre 1.x.x version](https://github.com/TryGhost/gscan/commit/99a4c5ddfa8a1c605e11be77f96d467866de2fa4) of gscan (good 'ol days). 

The change makes sense as is. The comments left are more of a nice to have to have to keep consistent style going forward 👍

PR #482, opened 2023-07-02
Removed unused bluebird Promise import from read-zip.js

vershwal:
Removed unused bluebird Promise import from read-zip.js for https://github.com/TryGhost/Ghost/issues/14882


daniellockyer:
Thanks @vershwal! See you around soon 🙂

PR #524, opened 2024-03-02
Fix incorrect "unused custom variables" error

### LINKED ISSUE #523: GS100-NO-UNUSED-CUSTOM-THEME-SETTING ignores partials
The "Custom theme settings defined in `package.json` must be used at least once in the theme templates" test explicitly ignores them files located in `/partials`, including [`/partials/content-cta.hbs`](https://ghost.org/docs/themes/members/#default-cta).

One of the [Ghost Docs examples](https://ghost.org/docs/themes/custom-settings/#text) demonstrates using custom variables to make the CTA text user-editable, but [doing so on my theme](https://github.com/yescomment/sl-sludgeline/commit/0c65c110c873dfb1b7d65e8e89fb7e406974293a) is now causing it to fail validation because `gscan` is not seeing the CTA variables used.

Why does gscan ignore `partials/*`? Should custom variables related to CTA be used somewhere other than `partials/content-cta` and imported?

unitof:
Closes #523

Proposing this as a quick fix as I don't see why hbs files in `partials/` are ignored for this test.

See #523 for specific examples.

However open to other approaches if there's a reason `partials/**/*.hbs` is ignored!

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -13,7 +13,7 @@ jobs:
       - uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4
         with:
           fetch-depth: 0
-      - uses: actions/setup-node@v3
+      - uses: actions/setup-node@v4
         env:
           FORCE_COLOR: 0
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,6 @@ on:
       - 'renovate/*'
 jobs:
   test:
-    uses: tryghost/actions/.github/workflows/test.yml@main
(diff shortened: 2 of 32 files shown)


unitof:
Nudging for feedback on this.

unitof:
@sagzy could you take a look at this? Would really help us get rid of the red error box on our admin, for using a custom variable in our CTA partial (an example from Ghost's own documentation).

We're on Ghost(Pro) so I can't use a gscan fork.

cathysarisky (approved):
I came here to put in the exact same PR.  Bump!  A merge would be helpful here. :)

allouis:
This appears to have been fixed here https://github.com/TryGhost/gscan/commit/404c80f4b4e357320c659c81d23f8b243b4e7287

Issue #464, opened 2023-01-19
Gscan does not detect custom setting usage inside a get helper


truekasun:
**featured.hbs**
![image](https://user-images.githubusercontent.com/278469/213362822-d73d1509-6dca-40b8-ab50-80f561b8e6d6.png)

**package.json**
![image](https://user-images.githubusercontent.com/278469/213362973-4bdba9a5-c6b3-4ef1-bf03-3b262416a4b5.png)

**gscan result on Ghost**
![image](https://user-images.githubusercontent.com/278469/213362530-23859866-dff6-4d6c-be47-413397ec7353.png)

---

**Note:**
The expected functionality (Tags filtered from the slugs inside the custom setting) is working fine. But it throws an error while validating.

ErisDS:
Hey there, thank you so much for the detailed bug report.

That does look like something that shouldn't happen! A PR to fix this issue would be very welcome 🙂

unitof:
@truekasun Is your `featured.hbs` located in your theme's `partials/` directory? If so, this may actually be #523, and not have anything to do with the `#get` helper function.

cathysarisky:
This is a separate bug from the partials/ problem.  (Although it might have BOTH problems.)  (I applied @unitof 's fix, and nope, these are still not parsed correctly.)

cathysarisky:
Adding on a bit with another example:
```{{#get "posts" filter="featured:false+tags:-[{{@custom.tag_1_slug}}]" limit=@custom.latest_post_count include="tags,authors" }}```

@custom.latest_post_count is OK.  But @custom.tag_1_slug is not.

PR #560, opened 2024-10-23
filter bug ugly hack

cathysarisky:
ref #464 

This is an ugly hack for failure of the parser to recognize custom variables when included in complex filter expression.

There's probably a great way to use the AST parser to do this.  I don't know what it is.

Tests pass.  Will also commit (coming shortly!) a set of tests that cover this use of custom variables.

### DIFF
--- a/lib/checks/100-custom-template-settings-usage.js
+++ b/lib/checks/100-custom-template-settings-usage.js
@@ -137,8 +137,7 @@ const ruleImplementations = {
             result.customThemeSettings = new Set();
         },
         eachFile: ({file, theme, log, result, partialVerificationCache}) => {
-            let templateTest = file.file.match(/(?<!partials\/.+?)\.hbs$/);
-
+            let templateTest = file.file.match(/\.hbs$/);
             if (templateTest) {
                 parseWithAST({file, theme, rules: {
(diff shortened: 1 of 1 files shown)


allouis:
Hey @cathysarisky it would be great if you could add a test which shows an example of the current code failing - just so we can understand the problem better - from your issue description I think it's when using the `get` helper with a dynamic filter property - is that right? Something like this? `{{#get "posts" filter="blah:{{@custom.thing}}"}}`

I think introducing regex parsing to an AST based checker is a step backwards and I'd like to see this shipped using the AST - I think it's doable!

cathysarisky:
Right, `{{#get "posts" filter="blah:{{@custom.thing}}"}}` is the problem case.  Struggled a bit with getting a test that worked.  Will pick this back up when I get a chance / am in the mood to bang my head on it a bit more.

Agreed, my approach is Wrong.

vikaspotluri123:
I think we need to reframe the issue to "#get.filter should be type-checked", since it's special and we interpolate static values as part of the helper. This would enable e.g. [no-unknown-globals](https://github.com/TryGhost/gscan/blob/ec2bc961a23adefe3fcdba91e7a946f9787dad3d/lib/ast-linter/rules/lint-no-unknown-globals.js) as well

Issue #573, opened 2025-05-03
A file with multiple "unknown custom theme setting errors" only shows one at a time


troyciesco:
## Issue summary
While learning how to build a theme, I was taking things from Casper and Source and editing them just to see how things work. The error messages in the sidebar were very helpful, but as I was going through and fixing them I noticed that if a file had a custom theme setting error it would only show the first one.

For example, in my index.hbs I have `@custom.demo_prop_on_index_1` and `@custom.demo_prop_on_index_2`. Neither actually exist in the config in `package.json`, so I'd expect to see an error for both. However, I just see the first one (same for the other files):

<img width="350" alt="Image" src="https://github.com/user-attachments/assets/fd6adff8-51f8-4440-9b5b-e4b4dec002a9" />

When I remove that piece of code and check the scan again, only now do I see the second error:

<img width="350" alt="Image" src="https://github.com/user-attachments/assets/1ab4eecb-db34-4884-8d83-24ce5ee6f59b" />

In addition, in a large file it can be a bit difficult to find where the error is because there's no line number.
## Steps to reproduce
- You'll need a theme with files that're using custom theme settings that aren't actually defined in package.json. I created an example here: https://github.com/troyciesco/ghost-skeleton and here's a zip: [ghost-skeleton.zip](https://github.com/user-attachments/files/20023890/ghost-skeleton.zip). Basically you want the file(s) to have multiple unknown custom theme setting errors.
- Check the scan using the web tool, directly in a ghost instance, or with `--verbose` in the CLI. You'll only see the first error for each file.
	- using the CLI without verbose will give you a comma delimited list of files with each file being listed only once, which i think makes sense.

## Proposed Solution (see PR)
- Display each error for each file, with a line number showing where the error happened (i.e. "L8")
- ensure the regular CLI version still displays a comma delimited list that only shows each file once

## Notes
- Similar things have been talked about a few times (https://github.com/TryGhost/gscan/issues/215 and to a lesser extent https://github.com/TryGhost/gscan/issues/122). I do like the idea of the file name being listed once and all the errors being nested underneath, but I figured this small improvement was enough to start with.

PR #574, opened 2025-05-03
🎨 Show all custom theme setting errors for each file with line number

troyciesco:
ref https://github.com/TryGhost/gscan/issues/573

- currently when a file has multiple custom theme setting errors, gscan only reports one at a time. This makes it difficult to debug a file if there are multiple errors.
- with this change, all errors for a file will be shown with a line number to make it easy to edit
- using the cli without the verbose tag still shows a comma-delimited list, with each file listed only once

before:
Screenshot 2025-05-03 at 10 42 21 AM

after (each file only has two errors, but i did test with 1 and 3):
Screenshot 2025-05-03 at 11 25 27 AM

cli output without `--verbose`:
```
Your theme has 1 error!
----

Errors
------
Important to fix, functionality may be degraded.

- Error: An unknown custom theme setting has been used.
Affected Files: index.hbs, partials/post-card.hbs, page.hbs, post.hbs

```

- The smaller footprint of this PR makes it easier to find and fix bugs in theme files without completely overhauling the checker

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -141,7 +141,7 @@ function outputResult(result, options) {
                 ui.log(message);
             });
         } else {
-            ui.log(`${chalk.bold('Affected Files:')} ${_.map(result.failures, 'ref')}`);
+            ui.log(`${chalk.bold('Affected Files:')} ${_.uniq(_.map(result.failures, 'ref')).join(', ')}`);
         }
     }
 
--- a/test/format.test.js
+++ b/test/format.test.js
@@ -535,7 +535,7 @@ describe('Format', function () {
 
                 theme.results.error.byFiles['assets/my.css'].length.should.eql(3);
                 theme.results.error.byFiles['default.hbs'].length.should.eql(21);
-                theme.results.error.byFiles['post.hbs'].length.should.eql(65);
+                theme.results.error.byFiles['post.hbs'].length.should.eql(66);
                 theme.results.error.byFiles['partials/mypartial.hbs'].length.should.eql(5);
                 theme.results.error.byFiles['index.hbs'].length.should.eql(14);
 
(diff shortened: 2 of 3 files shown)


test/format.test.js
```diff
@@ -535,7 +535,7 @@ describe('Format', function () {
 
                 theme.results.error.byFiles['assets/my.css'].length.should.eql(3);
                 theme.results.error.byFiles['default.hbs'].length.should.eql(21);
-                theme.results.error.byFiles['post.hbs'].length.should.eql(65);
+                theme.results.error.byFiles['post.hbs'].length.should.eql(66);
```
troyciesco, line 538: updating this test here makes sense, because before only `@price.monthly` was logged, but now `@price.yearly` is as well. File here: https://github.com/TryGhost/gscan/blob/main/test/fixtures/themes/001-deprecations/v5/invalid_all/post.hbs#L5

I don't *think* there are additional tests to add but I can if needed!

bin/cli.js
```diff
@@ -141,7 +141,7 @@ function outputResult(result, options) {
                 ui.log(message);
             });
         } else {
-            ui.log(`${chalk.bold('Affected Files:')} ${_.map(result.failures, 'ref')}`);
+            ui.log(`${chalk.bold('Affected Files:')} ${_.uniq(_.map(result.failures, 'ref')).join(', ')}`);
```
troyciesco, line 144: I prefer this output:
```
Affected Files: index.hbs, partials/post-card.hbs, page.hbs, post.hbs
```

to this:
```
Affected Files: index.hbs,partials/post-card.hbs,page.hbs,post.hbs
```
but if there was a reason for leaving out the spaces i can certainly switch it back. 

I do think `_.uniq` is useful, because otherwise it looks like this:
```
Affected Files: index.hbs,index.hbs,partials/post-card.hbs,partials/post-card.hbs,page.hbs,page.hbs,post.hbs,post.hbs
```

Although, now I'm wondering if there's ever a reason you'd want duplicated filenames for different failure results that are output by this function - probably not? But can change if needed!

PR #585, opened 2025-06-19
Optimized npm installation size

vikaspotluri123:
merge strategy: rebase

- created a separate package.json for app dependencies
- removed `uuid` in favor of `crypto.randomUUID`
- added explicit dependency on handlebars (previously came from express-hbs)
- node_modules went from ~168MB to ~72MB

Manual testing: Ran the app and uploaded a fixture, getting proper results.
Missing tests: Didn't test with Sentry, or that the Dockerfile works
Also: Not sure if the Renovate config needs to be picked up, or if it will automatically find app/package.json


JohnONolan:
stale / too difficult to reason about at this point - apologies for not getting to it in time

PR #609, opened 2025-08-11
🎨 Added deprecation warnings for facebook and twitter helpers

troyciesco:
ref https://linear.app/ghost/issue/PROD-1566

- adds a deprecation rule if a theme tries to use the facebook or twitter url helpers, and instructs the user to use the social_url helper instead

### DIFF
--- a/test/001-deprecations.test.js
+++ b/test/001-deprecations.test.js
@@ -3583,7 +3583,7 @@ describe('001 Deprecations', function () {
                 output.results.fail['GS001-DEPR-SITE-LANG'].should.be.a.ValidFailObject();
                 output.results.fail['GS001-DEPR-SITE-LANG'].failures.length.should.eql(1);
 
-                output.results.pass.should.be.an.Array().with.lengthOf(48);
+                output.results.pass.should.be.an.Array().with.lengthOf(50);
 
                 done();
             }).catch(done);
@@ -3677,12 +3677,12 @@ describe('001 Deprecations', function () {
 
                 output.results.fail['GS001-DEPR-AMP-TEMPLATE'].should.be.a.ValidFailObject();
                 output.results.fail['GS001-DEPR-AMP-TEMPLATE'].failures.length.should.eql(4);
-                
+
                 // Check all AMP template files are detected
                 const ampFiles = output.results.fail['GS001-DEPR-AMP-TEMPLATE'].failures.map(f => f.ref).sort();
                 ampFiles.should.eql([
                     'amp-lightning-with-attrs.hbs',
-                    'amp-lightning.hbs', 
+                    'amp-lightning.hbs',
                     'amp-with-class.hbs',
                     'amp.hbs'
                 ]);


test/001-deprecations.test.js
```diff
@@ -3702,5 +3702,37 @@ describe('001 Deprecations', function () {
...
+                );
+
+                output.results.fail['GS001-DEPR-FACEBOOK-URL'].should.be.a.ValidFailObject();
+                output.results.fail['GS001-DEPR-FACEBOOK-URL'].failures.length.should.eql(3);
+
+                output.results.fail['GS001-DEPR-TWITTER-URL'].should.be.a.ValidFailObject();
+                output.results.fail['GS001-DEPR-TWITTER-URL'].failures.length.should.eql(3);
+
+                const facebookFiles = output.results.fail['GS001-DEPR-FACEBOOK-URL'].failures.map(f => f.ref).sort();
+                facebookFiles.should.eql([
```
troyciesco, line 3722: i think i need each of these files to ensure the three different ways for using the helpers get caught in the check, but def open to doing it differently if there's a better way.

lib/specs/v5.js
```diff
@@ -707,6 +707,24 @@ let rules = {
         helper: '{{[#].currency_symbol}}',
         regex: /currency_symbol/g
     },
+    'GS001-DEPR-TWITTER-URL': {
```
troyciesco, line 710: added for the v5 checker since that's when social_url was actually introduced. 

though, we wouldn't completely remove the helpers until at least v7, and anyone still on v5 by then would have to go up to v6 first - so i could remove the v5 check.

lib/specs/v6.js
```diff
@@ -26,6 +26,24 @@ let rules = {
         Find more information about the <code>{{#get}}</code> helper <a href="${docsBaseUrl}helpers/get/" target=_blank>here</a>.`,
         helper: '{{#get}}'
     },
+    'GS001-DEPR-TWITTER-URL': {
+        level: 'error',
+        fatal: false,
```
troyciesco, line 31: It seems to me like a non-fatal error makes sense here, since once we (eventually) remove the fb/twitter helpers in a future major, then themes would break.

lib/specs/v5.js
kevinansfield: Keeping only v6 would be preferable I think as it's when we introduced the deprecation and keeps things cleaner. Also rules generally don't get duplicated across versions because each version merges all previous versions rules, the only time we duplicate is when a newer version overrides a rule in an older version to change properties.

Other than v5 being selected in the dropdown on the gscan website I don't think there's any situation where a theme would be tested explicitly against the v5 ruleset now 🤔 I don't think we'd go back to bump gscan version in the 5.x maintenance branch.

test/001-deprecations.test.js
kevinansfield: This approach makes sense, but I wonder if we need the earlier `.failures.length.should.eql(3);` assertions in this case because if one of these files didn't get picked up then that earlier assertion would fail and you'd never see the more helpful output from these individual file assertions anyway.

troyciesco:
Removed the ones I added to v5, but then realized the v6 test suite is referencing v5 in some cases (because nothing existed for v6 yet) so I'll review that and clean it up!

kevinansfield (approved):

PR #814, opened 2026-05-21
Rewrote the inline-dynamic-partial lint message to be actionable

ErisDS:
## Summary

The lint message that fires when a theme uses an inline dynamic partial (e.g. `{{> (concat "icons/" type)}}`) said *"...use block dynamic partials instead"* with no example. Theme authors who hit it had no way to know what "block dynamic partials" looked like or that Handlebars requires the closing tag to be `{{/undefined}}` — a parser quirk discovered through trial and error.

This was the message that initially confused the `social_accounts` helper reporter on the [forum thread](https://forum.ghost.org/t/social-accounts-v6-38-0-helper-for-theme-social-rendering/62854).

## Change

`lib/ast-linter/rules/lint-no-unknown-partials.js` — the message now leads with the fix shape, names the failure mode, and explicitly flags the `{{/undefined}}` close tag:

> Use the block form for dynamic partials so the page survives a missing partial: `{{#> (dynamicPartial)}}fallback markup{{/undefined}}` instead of `{{> (dynamicPartial)}}`. The inline form throws a page error when the named partial does not exist. Note the closing tag must be `{{/undefined}}`.

## Test plan

- [x] `yarn test` — 409 passing (updated the one substring assertion in `test/005-template-compile.test.js`)
- [x] `yarn lint` — clean

### DIFF
--- a/lib/ast-linter/rules/lint-no-unknown-partials.js
+++ b/lib/ast-linter/rules/lint-no-unknown-partials.js
@@ -19,7 +19,7 @@ module.exports = class NoUnknownPartials extends Rule {
                     }
                     if (node.type === 'PartialStatement') {
                         return this.log({
-                            message: `Use the block form for dynamic partials so the page survives a missing partial: <code>{{#> (dynamicPartial)}}fallback markup{{/undefined}}</code> instead of <code>{{> (dynamicPartial)}}</code>. See the <a href="https://ghost.org/docs/themes/structure/" target=_blank>partials documentation</a> for details.`,
+                            message: `Use the block form for dynamic partials so the page survives a missing partial: {{#> (dynamicPartial)}}fallback markup{{/undefined}} instead of {{> (dynamicPartial)}}.`,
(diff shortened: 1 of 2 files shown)


ErisDS:
Closing in favour of #815, which solves the same problem structurally:

- New dedicated `GS005-NO-INLINE-DYNAMIC-PARTIAL` spec rule with the correct title and severity, instead of piggybacking on `GS005-TPL-ERR` ("Templates must contain valid Handlebars") which was the wrong category — the template compiles fine.
- Failure message now includes the line number and exact source snippet (e.g. `Inline dynamic partial on line 35: {{> (concat "icons/social/" type)}}`).
- Side fix: `005-template-compile.js` refactored to run the linter once and route failures by code, removing a latent bug where adding any second GS005-* rule would have caused duplicate failures.

Benchmarked: no meaningful perf difference between main, #814, and #815 (trimmed mean within 0.4 ms across 200 iterations on Casper).

PR #816, opened 2026-05-21
Added recursive layout detection

sagzy:
ref https://linear.app/ghost/issue/BER-3133

## Problem
Users have caused site outages by introducing infinite recursion in their custom theme templates (specifically, using `{{!< default}}` in `default.hbs`).

##  Solution

Added a v6-only `GS130-NO-RECURSIVE-LAYOUT` rule to catch recursive `{{!< ...}}` template inheritance.

#### Error message:
CleanShot 2026-06-04 at 13 43 21@2x


lib/checks/130-template-inheritance.js
```diff
@@ -0,0 +1,156 @@
...
+
+function getLayoutReferences(themeFile) {
+    const source = normalizePath(themeFile.normalizedFile || themeFile.file);
+    const content = themeFile.content || '';
+    const references = [];
+    let match;
+
+    layoutPattern.lastIndex = 0;
+
+    while ((match = layoutPattern.exec(content)) !== null) {
```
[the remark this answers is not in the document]
sagzy: thanks, addressed in 4ea22b9ccc73b4b9ff56e37fd0a279cbf2db1b13

lib/checks/130-template-inheritance.js
```diff
@@ -0,0 +1,150 @@
...
+
+function ensureExtension(layoutPath) {
+    if (path.posix.extname(layoutPath)) {
+        return layoutPath;
+    }
+
+    return `${layoutPath}.hbs`;
+}
+
+function resolveLayoutPath(sourceFile, layoutName) {
```
troyciesco, line 17: One _possible_ issue here:

Ghost uses `express-hbs` without configuring `layoutsDir`, so declared layouts resolve relative to the file declaring them, not relative to the theme root.

For example:

```text
theme/
├── default.hbs
└── members/
    ├── base.hbs
    └── default.hbs
```

If `members/base.hbs` contains:

```hbs
{{!< default}}
```

Ghost resolves that to:

```text
members/default.hbs
```

The current implementation appears to resolve it to:

```text
default.hbs
```

So if there's a theme set up this way we might end up with false positives. _However_, i had an agent try to find a ghost theme that has a structure like this and (it claimed to) check like 450 of them and didn't find this in the wild lol. So it might not be a real issue. 

I don't think it'd happen with `default.hbs` either, but maybe if someone built a theme with a bunch of `layout.hbs` at every level or something? i'm not sure.

[I think this will fix it](https://github.com/TryGhost/gscan/pull/832) and the tests might explain the issue a little better, though i'm not 100% sure if i'm missing something. 

I made it as a PR into your branch here, so if you like it you can add it, if not we can merge your change as-is!

troyciesco (approved):

lib/checks/130-template-inheritance.js
sagzy: @troyciesco this is a great catch! Thanks a lot for the thorough review, I didn't expect you to open a PR with the fix and have an agent go through existing Ghost themes to double-check 🙏 🙏

PR #876, opened 2026-07-22
Allow color functions as default site background

samlfair:
This fix addresses https://github.com/TryGhost/Ghost/issues/22943

I updated the check for custom config colors.

Previously, the check tested for a hex code.

Now the check uses regex to check for hsl, hwb, rgb, lab, oklab, lch, oklch.

The check allows positive and negative numbers for all values, which may not be valid CSS.

The check allows for angles (rad, grad, turn, deg) or percents where valid.

The check does not allow for alphas.

The check does not allow for the keyword `none` or any special functions. 

This is the logic I used to test the regex:

```js
import { faker } from '@faker-js/faker'

const syntaxes = [
  /^(hsl|hwb)\(\-?\d+(\.\d+)?(deg|grad|rad|turn)? \-?\d+(\.\d+)?% \-?\d+(\.\d+)?%\)$/i,
  /^(rgb|lab|oklab)\(\-?\d+(\.\d+)?%?,? \-?\d+(\.\d+)?%?,? \-?\d+(\.\d+)?%?\)$/i,
  /^(lch|oklch)\(\-?\d+(\.\d+)?%? \-?\d+(\.\d+)?%? \-?\d+(\.\d+)?(deg|grad|rad|turn)?\)$/i,
  /^#[0-9a-f]{6}$/i,
  /^#[0-9a-f]{3}$/i
]

console.log(syntaxes.some(s => s.test("oklch(1 0 0)")))


const colorFunctions = ["rgb", "hsl", "hwb", "lab", "lch", "hex"]

const failures = []

colorFunctions.map(colorFunction => {
  for (let i = 0; i < 100; i++) {
    const color = colorFunction === "hex"
      ? faker.color.rgb({ format: "hex" })
      : faker.color[colorFunction]({ format: "css" })

    const valid = syntaxes.some(syntax => color.match(syntax))

    if (!valid) {
      const failure = { [colorFunction]: color }
      failures.push(failure)
    }
  }
})

const fakerModules = Object.keys(faker)

const successes = fakerModules.flatMap(key => {
  const fakerModule = faker[key]

  if(key === "color") return
  const fakerFunctions = Object.values(fakerModule)

  return fakerFunctions.map(fakerFunction => {
    if (typeof fakerFunction === "function") {
      try {
        const test = fakerFunction()
        if (typeof test === "string") {
          const valid = syntaxes.some(syntax => test.match(syntax))

          if(valid) {
            return { fakerModule: key, func: fakerFunction.name, fakeData: test }
          }
        }
      } catch (e) {
        null
      }
    }
  })
}).filter(a => a)

if (failures.length) {
  const plural = failures.length === 1 ? "failure" : "failures"
  console.log(`${failures.length} ${plural}:`)
  console.log(failures)
}

if (successes.length) {
  const plural = successes.length === 1 ? "test" : "tests"
  console.log(`${successes.length} random ${plural} passed erroneously:`)
  console.log(successes)
}
```

### DIFF
--- a/lib/checks/010-package-json.js
+++ b/lib/checks/010-package-json.js
@@ -195,10 +195,17 @@ _private.validatePackageJSONFields = function validatePackageJSONFields(packageJ
                 }
                 break;
             case 'color':
-                if (!/^#[0-9a-f]{6}$/i.test(entry.default)) {
-                    markFailed('invalidCustomThemeSetingColorDefault');
+                const syntaxes = [
+                  /^(hsl|hwb)\(\-?\d+(\.\d+)?(deg|grad|rad|turn)? \-?\d+(\.\d+)?% \-?\d+(\.\d+)?%\)$/i,
+                  /^(rgb|lab|oklab)\(\-?\d+(\.\d+)?%?,? \-?\d+(\.\d+)?%?,? \-?\d+(\.\d+)?%?\)$/i,
+                  /^(lch|oklch)\(\-?\d+(\.\d+)?%? \-?\d+(\.\d+)?%? \-?\d+(\.\d+)?(deg|grad|rad|turn)?\)$/i,
+                  /^#[0-9a-f]{6}$/i,
+                  /^#[0-9a-f]{3}$/i
+                ]
+
+                if (!syntaxes.some(s => s.test(entry.default))) {
+                  markFailed('invalidCustomThemeSetingColorDefault');
                 }
-                break;
             case 'image':
                 if (entry.default) {
                     markFailed('invalidCustomThemeSetingImageDefault');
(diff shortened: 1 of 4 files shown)


troyciesco:
Thanks for your effort here, but we'll actually have to make a handful of changes to Ghost before gscan for this to work properly.

For example, if i set my theme's default `site_background_color` to `hsl(170, 85%, 52%)`, the color picker in the Design & Branding area of settings won't work properly. the color picker will show `#ffffff`, though the picker value is empty (because it didn't process the hsl value):
image

This also causes validation errors in that modal:
image

(Worth mentioning that in general, the picker does let a user _input_ `hsl`, but it gets converted back to hex before saving.)

I'm not sure if there are other areas where this might break, but that was a concrete one I found.

I'm going to close this for now to avoid confusion, but it's a useful reference if we pick up the larger changes in Ghost.

PR #891, opened 2026-08-15
Fix false-positive {{author}} deprecation warning inside {{#is "author"}} blocks

### LINKED ISSUE #365: Gscan complains about correct usage of `{{author}}` in context of `default.hbs`
This is the logic of my theme's page title in `default.hbs` which actually **works**:

```hbs
    <title>
        {{#is "home"}}
            {{@site.title}} — {{@site.description}}
        {{/is}}
        {{#is "author"}}
            {{author.name}} | {{@site.title}}
        {{/is}}
        {{#is "tag"}}
            {{tag.name}} | {{@site.title}}
        {{/is}}
        {{^is "home, author, tag"}}
            {{meta_title}} | {{@site.title}}
        {{/is}}
    </title>
```

However, when I run `gcan` (the latest version, 3.6.0), it throws an error saying:

```
Checking theme compatibility...

Your theme has 1 error!
----

Errors
------
Important to fix, functionality may be degraded.

- Error: Replace the {{author.name}} helper with {{primary_author.name}} or {{authors.[#].name}}
Files: default.hbs
```

I think this is a false alarm which should be fixed (I wish I could do it).

wakqasahmed:
Fixes #365

## Summary

`gscan` incorrectly flags `{{author.name}}` (and other `{{author.*}}` property accesses) as a deprecated helper usage even when it appears inside an `{{#is "author"}}...{{/is}}` block, e.g. in `default.hbs`:

```hbs
{{#is "author"}}
    {{author.name}} | {{@site.title}}
{{/is}}
```

This is valid, documented usage — inside an `{{#is "author"}}` block the current context genuinely is the Author object, so `{{author.name}}` (unlike `{{author.name}}` used elsewhere, e.g. inside a post loop) is correct and should not trigger the "Replace `{{author.name}}` with `{{primary_author.name}}` or `{{authors.[#].name}}`" deprecation error.

The regex-based `GS001-DEPR-AUTH-*` deprecation checks in `lib/checks/001-deprecations.js` didn't have any context awareness for this, unlike the `{{#author}}` block-helper rule which already excludes `author.hbs` via `notValidIn`.

## Fix

`lib/checks/001-deprecations.js` now strips out the contents of `{{#is "..."}}...{{/is}}` blocks that include `author` in their condition list before running the `GS001-DEPR-AUTH*` regex checks, so usage of `{{author}}`/`{{author.*}}` inside those blocks is no longer flagged, in any template (not just `author.hbs`).

## Tests

- Added a new fixture theme (`test/fixtures/themes/001-deprecations/v6/valid-author-is-context/default.hbs`) reproducing the exact `default.hbs` snippet from the issue.
- Added a regression test in `test/001-deprecations.test.js` asserting no failures are reported for that fixture.
- Verified the new test fails without the fix (reproducing the reported false positive) and passes with it.
- Ran the full `001-deprecations` test suite (29/29 passing) and the full test suite (pre-existing, unrelated failure in `test/general.test.js` around `Thumbs.db` handling reproduces identically on `main` without this change).
- `eslint` clean on changed files.

### Test plan
- [x] `npx vitest run test/001-deprecations.test.js` — 29/29 passing
- [x] `npx vitest run` — same pass/fail counts as on `main` (1 pre-existing, unrelated failure)
- [x] `npx eslint lib/checks/001-deprecations.js test/001-deprecations.test.js` — clean

### DIFF
--- a/lib/checks/001-deprecations.js
+++ b/lib/checks/001-deprecations.js
@@ -12,10 +12,14 @@ const versions = require('../utils').versions;
 // discussion for the concrete false-negative cases a naive strip previously
 // introduced.
 //
+// `{{#if}}`/`{{#unless}}` are tracked as their own frame type so that a
+// nested `{{else}}` (which belongs to that if/unless, not the enclosing
+// `{{#is "author"}}`) never flips the outer is-block's active branch.
+//
 // Note: bare `{{author}}` (GS001-DEPR-AUTH) is intentionally NOT in the list
 // below - it's a removed helper, not a property access, so it renders
 // `[object Object]` even inside an author context and should never be exempt.
-const authorTagRegex = /{{!--[\s\S]*?--}}|{{!(?!--)[\s\S]*?}}|{{[#^]is\s+(['"])([\s\S]*?)\1\s*}}|{{\/is}}|{{else}}|{{#(?:foreach|each)\b[^}]*}}|{{\/(?:foreach|each)}}/g;
+const authorTagRegex = /{{!--[\s\S]*?--}}|{{!(?!--)[\s\S]*?}}|{{[#^]is\s+(['"])([\s\S]*?)\1\s*}}|{{\/is}}|{{else}}|{{#(?:foreach|each)\b[^}]*}}|{{\/(?:foreach|each)}}|{{[#^](?:if|unless)\b[^}]*}}|{{\/(?:if|unless)}}/g;
 
 function currentExempt(stack) {
     if (!stack.length) {
@@ -62,6 +66,10 @@ function getAuthorCheckableContent(content) {
                 elseExempt: ownElseExempt || exemptBefore
             });
         } else if (token === '{{else}}') {
+            // Only an `{{#is}}` frame's own `{{else}}` should switch its
+            // active branch - an `{{else}}` belonging to a nested
+            // `{{#if}}`/`{{#unless}}` (or `{{#foreach}}`/`{{#each}}`) frame
+            // must not leak through to the enclosing `is` frame.
             if (stack.length && stack[stack.length - 1].type === 'is') {
                 stack[stack.length - 1].inElse = true;
             }
@@ -76,6 +84,19 @@ function getAuthorCheckableContent(content) {
             stack.push({type: 'loop'});
         } else if (/^{{\/(?:foreach|each)}}/.test(token)) {
             if (stack.length && stack[stack.length - 1].type === 'loop') {
+                stack.pop();
+            }
+        } else if (/^{{[#^](?:if|unless)\b/.test(token)) {
+            // `{{#if}}`/`{{#unless}}` don't change author context - both of
+            // their branches inherit whatever exemption was already active.
+            stack.push({
+                type: 'cond',
+                inElse: false,
+                primaryExempt: exemptBefore,
+                elseExempt: exemptBefore
+            });
+        } else if (/^{{\/(?:if|unless)}}/.test(token)) {
+            if (stack.length && stack[stack.length - 1].type === 'cond') {
                 stack.pop();
             }
         }
--- a/lib/checks/001-deprecations.js
+++ b/lib/checks/001-deprecations.js
@@ -7,14 +7,29 @@ const versions = require('../utils').versions;
 // default.hbs), so deprecation checks for those properties should ignore
 // content in these blocks. Tokenizes rather than doing a single blind regex
 // replace so that: multi-condition `{{#is}}` (an OR, so it isn't necessarily
-// author context), Handlebars comments, and nested `{{#is}}`/`{{#foreach}}`
-// blocks are all handled correctly - see #365 review discussion for the
-// concrete false-negative cases a naive strip previously introduced.
+// author context), Handlebars comments, nested `{{#is}}`/`{{#foreach}}`
+// blocks, and `{{else}}` branches are all handled correctly - see #365 review
+// discussion for the concrete false-negative cases a naive strip previously
+// introduced.
 //
 // Note: bare `{{author}}` (GS001-DEPR-AUTH) is intentionally NOT in the list
 // below - it's a removed helper, not a property access, so it renders
 // `[object Object]` even inside an author context and should never be exempt.
-const authorTagRegex = /{{!--[\s\S]*?--}}|{{!(?!--)[\s\S]*?}}|{{[#^]is\s+(['"])([\s\S]*?)\1\s*}}|{{\/is}}|{{#(?:foreach|each)\b[^}]*}}|{{\/(?:foreach|each)}}/g;
+const authorTagRegex = /{{!--[\s\S]*?--}}|{{!(?!--)[\s\S]*?}}|{{[#^]is\s+(['"])([\s\S]*?)\1\s*}}|{{\/is}}|{{else}}|{{#(?:foreach|each)\b[^}]*}}|{{\/(?:foreach|each)}}/g;
+
+function currentExempt(stack) {
+    if (!stack.length) {
+        return false;
+    }
+
+    const top = stack[stack.length - 1];
+
+    if (top.type === 'loop') {
+        return false;
+    }
+
+    return top.inElse ? top.elseExempt : top.primaryExempt;
+}
 
 function getAuthorCheckableContent(content) {
     const stack = [];
@@ -24,7 +39,7 @@ function getAuthorCheckableContent(content) {
 
     authorTagRegex.lastIndex = 0;
     while ((match = authorTagRegex.exec(content)) !== null) {
-        const exemptBefore = stack.length ? stack[stack.length - 1].exempt : false;
+        const exemptBefore = currentExempt(stack);
         result += exemptBefore ? '' : content.slice(lastIndex, match.index);
 
         const token = match[0];
@@ -33,8 +48,23 @@ function getAuthorCheckableContent(content) {
             // Handlebars comment: opaque, does not affect nesting/exemption.
         } else if (token.startsWith('{{#is') || token.startsWith('{{^is')) {
             const condition = (match[2] || '').trim();
-            const isSingleAuthorBlock = token.startsWith('{{#is') && condition === 'author';
-            stack.push({type: 'is', exempt: isSingleAuthorBlock || exemptBefore});
+            const isNegated = token.startsWith('{{^is');
+            const isSingleAuthorCondition = condition === 'author';
+            // {{#is "author"}}: author context before {{else}}, not after.
+            // {{^is "author"}}: the inverse - author context only after
+            // {{else}} (Ghost's {{#is}} supports {{else}} for both forms).
+            const ownPrimaryExempt = isSingleAuthorCondition && !isNegated;
+            const ownElseExempt = isSingleAuthorCondition && isNegated;
+            stack.push({
+                type: 'is',
+                inElse: false,
+                primaryExempt: ownPrimaryExempt || exemptBefore,
+                elseExempt: ownElseExempt || exemptBefore
+            });
+        } else if (token === '{{else}}') {
+            if (stack.length && stack[stack.length - 1].type === 'is') {
+                stack[stack.length - 1].inElse = true;
+            }
         } else if (token === '{{/is}}') {
             if (stack.length && stack[stack.length - 1].type === 'is') {
                 stack.pop();
@@ -43,7 +73,7 @@ function getAuthorCheckableContent(content) {
             // {{author}} inside a {{#foreach}}/{{#each}} refers to that
             // item's author relation, not the page-context author, even
             // when nested inside an {{#is "author"}} block - still checkable.
-            stack.push({type: 'loop', exempt: false});
+            stack.push({type: 'loop'});
         } else if (/^{{\/(?:foreach|each)}}/.test(token)) {
             if (stack.length && stack[stack.length - 1].type === 'loop') {
                 stack.pop();
@@ -54,7 +84,7 @@ function getAuthorCheckableContent(content) {
         lastIndex = authorTagRegex.lastIndex;
     }
 
-    const exemptAfter = stack.length ? stack[stack.length - 1].exempt : false;
+    const exemptAfter = currentExempt(stack);
     result += exemptAfter ? '' : content.slice(lastIndex);
 
     return result;
--- a/lib/checks/001-deprecations.js
+++ b/lib/checks/001-deprecations.js
@@ -2,13 +2,65 @@ const _ = require('lodash');
 const spec = require('../specs');
 const versions = require('../utils').versions;
 
-// {{author}} and its properties are correctly available in the current context
-// while inside an {{#is "author"}}...{{/is}} block (e.g. in default.hbs), so
-// deprecation checks for those helpers should ignore content in these blocks.
-const authorIsBlockRegex = /{{#is\s+[^}]*\bauthor\b[^}]*}}[\s\S]*?{{\/is}}/g;
+// {{author.*}} properties are correctly available in the current context while
+// inside a single-condition {{#is "author"}}...{{/is}} block (e.g. in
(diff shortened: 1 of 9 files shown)


lib/checks/001-deprecations.js
```diff
@@ -2,6 +2,38 @@ const _ = require('lodash');
 const spec = require('../specs');
 const versions = require('../utils').versions;
 
+// {{author}} and its properties are correctly available in the current context
+// while inside an {{#is "author"}}...{{/is}} block (e.g. in default.hbs), so
+// deprecation checks for those helpers should ignore content in these blocks.
+const authorIsBlockRegex = /{{#is\s+[^}]*\bauthor\b[^}]*}}[\s\S]*?{{\/is}}/g;
```
wakqasahmed: Four concrete problems with this pattern, all reproduced end-to-end against `--v6`:

1. **Multi-context `{{#is}}` is an OR.** `[^}]*\bauthor\b[^}]*` matches `{{#is "index, author"}}` and `{{#is "post, author"}}`. Those blocks also render on index/post pages where `author` is not in scope, so stripping them silences a genuine error. `main` reports `GS001-DEPR-AUTH-NAME` for `{{#is "index, author"}}{{author.name}}{{/is}}`; this branch reports nothing.

2. **Non-greedy match breaks on nesting.** `[\s\S]*?{{\/is}}` stops at the first closing tag, so `{{#is "author"}}{{#is "paged"}}x{{/is}}{{author.name}}{{/is}}` leaves `{{author.name}}` outside the stripped range and still errors — the original false positive survives in nested templates. A balanced-scan (or AST walk) is needed.

3. **Comment-unaware.** `{{!-- {{#is "author"}} --}}` opens a match that runs to the next unrelated `{{/is}}`, hiding any real `{{author.*}}` in between. Verified: that input yields no error on this branch.

4. **Lexical containment != runtime context.** `{{#is "author"}}{{#foreach posts}}{{author.name}}{{/foreach}}{{/is}}` is post context at the point of use and is genuinely deprecated, but gets stripped. `GS090-NO-AUTHOR-HELPER-IN-POST-CONTEXT` does not catch it either, so this case goes from reported to entirely silent.

lib/checks/001-deprecations.js
```diff
@@ -2,6 +2,38 @@ const _ = require('lodash');
...
+
+const authorContextAwareRules = [
+    'GS001-DEPR-AUTH',
```
wakqasahmed: `GS001-DEPR-AUTH` matches bare `{{author}}` (`/{{\s*?author\s*?}}/`). That is a removed *helper*, not a property access — inside an `{{#is "author"}}` block `author` is an object, so `{{author}}` renders `[object Object]` and is still wrong. This entry masks a real breakage and should come out of the exemption list even if the rest of the approach is kept.

Same question for `GS001-DEPR-CON-AUTH` (`{{#if author.*}}`): worth stating explicitly in the PR description why each code in this list is context-dependent rather than unconditionally deprecated, since the list is what determines the blast radius.

lib/checks/001-deprecations.js
```diff
@@ -2,6 +2,38 @@ const _ = require('lodash');
...
+    'GS001-DEPR-AUTH-LOC',
+    'GS001-DEPR-AUTH-WEB',
+    'GS001-DEPR-AUTH-TW',
+    'GS001-DEPR-AUTH-FB',
+    'GS001-DEPR-AUTH-PIMG',
+    'GS001-DEPR-AUTH-CIMG',
+    'GS001-DEPR-AUTH-URL'
+];
+
+function getCheckableContent(content, ruleCode) {
```
wakqasahmed, line 132: Minor, secondary to the correctness issues: this runs a full-file `String#replace` for every (author rule x .hbs file) pair — 16 rules times every template on every scan, all producing the identical stripped string. If this shape survives review, hoist the stripped content to once per file (or memoise per `themeFile`) rather than recomputing inside the rule loop.

test/001-deprecations.test.js
```diff
@@ -3675,6 +3675,16 @@ describe('001 Deprecations', function () {
             });
         });
 
+        it('[success] should not flag {{author.*}} used inside an {{#is "author"}} block', function () {
```
wakqasahmed, line 3678: This locks in the fix but not its boundaries. Because `results.fail` is empty, *every* rule lands in `results.pass`, which makes the `assertContains(output.results.pass, 'GS001-DEPR-AUTH-NAME')` assertion pass trivially — it would pass even if the check had been disabled outright.

Please add the negative half so a future over-broad exemption gets caught: a fixture containing deprecated author usage **outside** the `{{#is "author"}}` block, plus the shapes that currently regress — `{{#is "index, author"}}{{author.name}}{{/is}}` and `{{#is "author"}}{{#foreach posts}}{{author.name}}{{/foreach}}{{/is}}` — asserting each is still reported. Both of those are reported on `main` and silent on this branch today.

wakqasahmed (commented):
Thanks for taking this on — the underlying complaint in #365 is real, the fixture reproduces it exactly (I confirmed the new test fails on `e4294bc` and passes on this branch, and `npx vitest run test/001-deprecations.test.js` is 29/29 green here), and the scoping to a named rule list is a sensible instinct.

However, I don't think the text-stripping approach can ship as-is: it silences several classes of **genuinely invalid** author usage that `main` currently reports. I reproduced each of these end-to-end with `./bin/cli.js <theme> --v6` on this branch vs. `main`.

### Confirmed false negatives introduced by this PR

All of these produce `Error: Replace {{author.name}} with ...` on `main`, and **no error at all** on this branch:

| # | Template snippet (in `index.hbs`) | Why it is still wrong |
|---|---|---|
| 1 | `{{#is "index, author"}}{{author.name}}{{/is}}` | `{{#is}}` with multiple contexts is an **OR**. On the index page `author` is not in scope, so this is exactly the deprecated usage the rule exists to catch — but `\bauthor\b` matches the argument list and the whole block is stripped. |
| 2 | `{{#is "post, author"}}{{author.id}}{{/is}}` | Same OR problem, post context. |
| 3 | `{{#is "author"}}{{#foreach posts}}{{author.name}}{{/foreach}}{{/is}}` | Inside `{{#foreach posts}}` the current context is a **post**, so `{{author.name}}` here resolves to the post's author and *is* deprecated. Being lexically inside an `{{#is "author"}}` block says nothing about the context at the point of use. `GS090-NO-AUTHOR-HELPER-IN-POST-CONTEXT` does not pick this one up either — I get a completely clean author section for it. |
| 4 | `{{!-- {{#is "author"}} --}}\n{{author.name}}\n{{#is "index"}}y{{/is}}` | The regex is comment-unaware, so a commented-out opening tag swallows everything up to the next unrelated `{{/is}}`, hiding the real `{{author.name}}` in between. |
| 5 | `{{#is "author"}}{{author}}{{/is}}` | `GS001-DEPR-AUTH` (bare `{{author}}`) is in the exemption list, but bare `{{author}}` is a removed *helper* — on an author page `author` is an object, so this renders `[object Object]` regardless of context. This one arguably should never have been exempted. |

### Known remaining false positive

`{{#is "author"}}{{#is "paged"}}x{{/is}}{{author.name}}{{/is}}` still errors — the non-greedy `[\s\S]*?` stops at the *first* `{{\/is}}`, so a nested block helper truncates the stripped region and the issue's own complaint reappears. So the fix is both too broad (above) and too narrow (here).

### Direction

This is precisely the case @naz called out on #365 ("a good example of why migrating to AST-based rules for author-related errors is a good idea", ref 377e46d). The repo already has the machinery: `lib/ast-linter/` with `internal/scope.js` and `lint-no-author-helper-in-post-page-context.js`, which tracks parent context via `this.scope.hasParentContext(...)`. An AST rule would get correct block nesting for free, could distinguish single-value from multi-value `{{#is}}` arguments, and would correctly re-flag usage inside a nested `{{#foreach posts}}`/`{{#post}}` frame. I'd suggest moving the author deprecation rules over there rather than layering a pre-processing pass on top of the regex engine — but that's a maintainer call, so please get sign-off on the approach before investing in the rewrite.

If the regex route is kept for now, at minimum it needs to: require the `{{#is}}` argument list to contain **only** `author`; be re-flagged inside any nested block that changes context; handle nested `{{#is}}` via a balanced scan rather than `*?`; skip `{{!-- --}}` and `{{!  }}` comments; and drop `GS001-DEPR-AUTH` from the exemption list.

### Smaller notes

- `{{#is "author"}}{{#author}}...{{/author}}{{/is}}` still reports `GS001-DEPR-AUTHBL`, which is inconsistent with the new exemption (that rule is only excused in `author.hbs` via `notValidIn`). Worth deciding deliberately either way.
- Test coverage: the new test only asserts `results.fail` is empty. Since an empty `fail` puts *every* rule into `pass`, the follow-up `assertContains(output.results.pass, 'GS001-DEPR-AUTH-NAME')` is close to vacuous. A behaviour-locking test needs the negative half: a fixture where deprecated author usage appears both inside an `{{#is "author"}}` block **and** outside it (and in the multi-value/nested-`foreach` shapes above), asserting the outside ones are still reported.
- Housekeeping: no CLA check ran on this PR and the repo has no `CONTRIBUTING.md`/PR template, so it's worth confirming with the maintainers whether Ghost's standard contributor agreement applies before merge. Informational only.

wakqasahmed:
Thanks for the thorough review — all 4 concrete false-negative cases plus the nesting issue were reproduced against current \`main\` and are real. Pushed a fix that replaces the blind regex-replace with a small tokenizer over \`{{#is}}\`/\`{{/is}}\`/\`{{#foreach}}\`/\`{{#each}}\`/Handlebars-comment tokens, tracking a nesting stack instead of a single non-greedy match.

**Holes closed:**
1. Multi-condition \`{{#is "index, author"}}\` / \`{{#is "post, author"}}\` — only a *single-condition* \`{{#is "author"}}\` now grants the exemption; multi-condition blocks are checked as before (regression added: \`multi-condition.hbs\`).
2. Nested \`{{#is}}\` blocks — the tokenizer tracks nesting depth via a stack instead of stopping at the first \`{{/is}}\`, so \`{{#is "author"}}{{#is "paged"}}x{{/is}}{{author.name}}{{/is}}\` is correctly still exempted (added to \`valid-author-is-context/default.hbs\`), and multi-condition/other nested \`{{#is}}\` blocks no longer break the boundary.
3. Handlebars comments — \`{{!-- ... --}}\` and \`{{! ... }}\` are now treated as opaque tokens that never open/close a real block, so a commented-out \`{{#is "author"}}\` can no longer swallow real, unrelated \`{{author.*}}\` usage after it (regression added: \`comment-not-a-block.hbs\`).
4. \`{{#foreach}}\`/\`{{#each}}\` nested inside \`{{#is "author"}}\` — the tokenizer forces content back to checkable inside a loop frame, since \`{{author.*}}\` there refers to that item's author relation, not the page-context author (regression added: \`foreach-shadow.hbs\`).
5. Bare \`{{author}}\` (\`GS001-DEPR-AUTH\`) — removed from the exemption list entirely. It's a removed helper, not a property access, and still renders \`[object Object]\` even inside a genuine author-is block (regression added: \`bare-author-helper.hbs\`).

**AST-based approach:** investigated reusing \`lib/ast-linter/rules/internal/scope.js\`'s \`hasParentContext()\`. It only tracks context via \`Frame\`, and \`getContext()\` for a generic \`BlockStatement\` is currently a stub (\`return {context: 'str', locals: {}, blockParams: {}}\`) — there's no existing logic resolving \`{{#is "author"}}\` into an author-context frame at all. Building that out properly (parsing \`is\` conditions, handling the OR semantics, wiring it into the existing post/page context frames) is a materially larger change than this fix, so per the guidance to prefer the narrower, more conservative fix, I kept this as a tokenizer-based regex fix rather than a partial/rushed AST migration. Flagging this as a good follow-up if the maintainers want the AST-linter to eventually own all author-context resolution (as referenced in naz's comment on #365).

**Non-blocking items:**
- Added a negative-case fixture/test (\`invalid-author-is-context-edge-cases\`) so the regression test can no longer pass trivially via an empty \`results.fail\`.
- Left the "recompute \`getCheckableContent\` once per rule" perf note as-is — it's now a tokenizer pass rather than a single regex replace, so the cost is slightly higher per call, but still only runs for the 15 author-context rules against \`.hbs\` files, and hoisting it out to a memoized per-file computation would require restructuring the check's rule-then-file iteration order. Happy to follow up separately if this is a measured concern.

Ran \`npx vitest run test/001-deprecations.test.js\` (30/30 passing) and the full \`npx vitest run\` (425/426 passing) — the one remaining failure is the pre-existing, unrelated \`Thumbs.db\` fixture-mutation issue in \`test/general.test.js\`, not something introduced here.

wakqasahmed:
Hi @coderabbitai — just checking in on this one, it's been about 4 days with CI green and mergeable. Is there anything needed from our side to help move it forward, or any rough idea on timing? No rush, just following up. Thanks for the review!

wakqasahmed:
Good catch — pushed a fix in 4cea5f4.

`getAuthorCheckableContent()` now tokenizes `{{else}}` as well: each `{{#is}}`/`{{^is}}` stack frame tracks a separate exempt state for its primary and else branches, and `{{else}}` flips which one is active for the innermost `is` block. So:

- `{{#is "author"}}...{{else}}...{{/is}}` — only the primary branch is author-exempt now; the else branch is checked as before.
- `{{^is "author"}}...{{else}}...{{/is}}` — the inverse: only the else branch is exempt.

Added `else-branch.hbs` covering both forms and asserted `GS001-DEPR-AUTH-MT`/`GS001-DEPR-AUTH-SLUG` still fail in the non-author branches while `GS001-DEPR-AUTH-WEB` stays exempt in the genuine author branch. `npx vitest run test/001-deprecations.test.js` is 30/30 passing, full suite 425/425, eslint clean.

lib/checks/001-deprecations.js
```diff
@@ -2,6 +2,120 @@ const _ = require('lodash');
...
+            stack.push({
+                type: 'is',
+                inElse: false,
+                primaryExempt: ownPrimaryExempt || exemptBefore,
+                elseExempt: ownElseExempt || exemptBefore
+            });
+        } else if (token === '{{else}}') {
+            if (stack.length && stack[stack.length - 1].type === 'is') {
+                stack[stack.length - 1].inElse = true;
+            }
```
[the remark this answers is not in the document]
wakqasahmed: Confirmed and fixed in 0708542.

`getAuthorCheckableContent()` now tracks `{{#if}}`/`{{#unless}}` as their own `cond` stack frame (both branches inherit whatever exemption was already active from the enclosing `is` frame), so a nested if/unless `{{else}}` only ever affects its own frame and can no longer flip the outer `is`-block's active branch.

Added `if-unless-nested.hbs`: `{{author.email}}`/`{{author.location}}` in the if/else branches of an `{{#if}}` nested inside `{{#is "author"}}` stay exempt (both were previously incorrectly flagged after the else toggled the outer frame), and `{{author.twitter}}` inside an unrelated top-level `{{#unless}}` still gets flagged. Verified the new assertions fail on the pre-fix code and pass on this commit. `npx vitest run test/001-deprecations.test.js` is 30/30, full suite 425/425, eslint clean.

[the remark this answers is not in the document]
wakqasahmed: Thanks for confirming — appreciate you re-verifying against the latest commit. Looks like the thread ended up marked resolved on GitHub's side after all, so no further action needed here.

wakqasahmed:
Hi @coderabbitai — following up again — still green and mergeable, about 4 days since the last activity. Happy to change anything that's blocking it, otherwise just don't want it lost. Thanks!

troyciesco:
Closing this for two reasons:

- There is a ton of agent-generated conversation, comments, etc. in this PR and it's too much to review closely
- If the AST approach is actually the correct way to do it, it doesn't make sense to merge this and have to remove later

You're welcome to submit a PR for the AST approach if that interests you, but please ensure the PR doesn't have an overwhelming amount of agent commentary so maintainer review can focus on the actual change.

PR #908, opened 2026-09-03
fix: rewrite author-context exemption using the Handlebars AST

wakqasahmed:
Replaces #891, which @troyciesco closed - the regex-tokenizer approach in that PR had too much back-and-forth to review, and an AST-based rewrite made more sense than merging something that'd need redoing anyway. This is that rewrite.

`getAuthorCheckableContent()` in `001-deprecations.js` used to simulate Handlebars block nesting by hand with a regex scan and a manual stack. It now parses the template with `lib/ast-linter/parser.js` (already used elsewhere in this repo) and walks the real AST to find which text/mustache ranges fall inside an `{{#is "author"}}` (or negated `{{^is}}`) block. Nesting is structural in the AST, so this handles `{{#if}}`/`{{#unless}}`, `{{#foreach}}`/`{{#each}}` shadowing, `{{else}}` branches, comments that just mention `{{#is "author"}}` in passing, and multi-condition `{{#is "a, b"}}` without any manual state tracking. Falls back to the original content if a template fails to parse.

Same behavior as before - verified against the full fixture set from #891's review (multi-condition, else-branch, foreach-shadow, comment-not-a-block, bare-author-helper, if-unless-nested), all still passing. `authorTagRegex` and the stack-tracking helpers are gone since nothing uses them anymore.

### DIFF
--- a/lib/checks/001-deprecations.js
+++ b/lib/checks/001-deprecations.js
@@ -67,7 +67,6 @@ function walkProgram(program, exempt, inLoop, content, lineOffsets, ranges) {
         }
 
         const isLoop = node.path.original === 'foreach' || node.path.original === 'each';
-        const isCondition = node.path.original === 'is';
         const isAuthorCondition = isSingleAuthorCondition(node);
         if (isAuthorCondition) {
             const isNegated = isNegatedBlock(content, node, lineOffsets);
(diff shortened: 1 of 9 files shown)


wakqasahmed:
Confirmed the CodeRabbit finding was real - wrote a repro with `{{#is "author"}}{{#is "paged"}}{{author.name}}{{/is}}{{/is}}` and the nested is block was clearing the exempt flag, so author.name inside it got flagged as a deprecation even though it's still in author context. Pushed a fix that only clears exempt when entering a loop, propagating it unchanged through both branches of a nested is. Added an author.name reference inside the nested paged block in the existing valid-author-is-context fixture and confirmed it goes red without the fix and green with it. Full suite (425 tests) still passes.

PR #917, opened 2026-09-03
Migrate from the glob package to Node's native fs.glob

acburdine:
## What & why

Node's `fsPromises.glob`/`fs.globSync` became stable (no longer behind an experimental flag/warning) in v22.17.0. Bumps the `engines.node` floor to `^22.17.0 || ^24.0.0` and moves both of gscan's `glob` usages to the built-in, dropping the `glob` dependency.

Split out as its own PR (ahead of #909) per review discussion there - the Node-floor bump and dependency replacement stand alone, then the `readThemeStructure` bounded-concurrency rewrite (addressing the unbounded directory-read fan-out flagged on #909) lands on top of this.

## Changes

- `engines.node` bumped, CI matrix's pinned 22.x version bumped to match
- `lib/read-zip.js`: `glob('**/index.hbs', {cwd, nosort: true})` → `fs.glob()` + `Array.fromAsync`. The old call already made no ordering guarantee (`nosort: true`); for the edge case of a malformed zip with more than one `index.hbs`, matches are now sorted by path depth (then alphabetically) and the shallowest wins - deterministic where before it was whatever order the old library happened to return.
- `test/valid-doc-links.js`: `glob.sync('*.js', {ignore: ...})` → `fs.globSync()` + a manual filter (the native `exclude` option didn't behave as documented in local testing).
- `glob` removed from `dependencies`; `yarn.lock` updated.
- `resolveBaseDir` exposed via `module.exports._private` in `read-zip.js`, with new unit tests for the depth-sort and its tiebreak (no existing fixture covered the multi-match case).

## Testing

- `yarn test` (437 tests, all green)
- `yarn lint` clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [22.13.1, 24]
+        node-version: [22.17.0, 24]
     env:
       CI: true
       FORCE_COLOR: 1
--- a/lib/read-zip.js
+++ b/lib/read-zip.js
@@ -1,8 +1,8 @@
 const debug = require('@tryghost/debug')('zip');
 const path = require('path');
 const os = require('os');
+const fs = require('fs/promises');
 const {randomUUID} = require('crypto');
-const {glob} = require('glob');
 const {extract} = require('@tryghost/zip');
 const errors = require('@tryghost/errors');
 const _ = require('lodash');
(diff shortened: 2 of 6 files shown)


acburdine:
Review finding (P2): the new shallowest-match selection is not path-separator portable.

`a.split("/").length` assumes `fs.glob()` returns POSIX-separated paths. Native filesystem paths on Windows may use `\\`, in which case both `deep\\nested\\index.hbs` and `shallow\\index.hbs` have a computed depth of 1. The alphabetical tiebreak can then select the deeper theme root, defeating the stated shallowest-first behavior. The final `.replace(/\\/$/, "")` similarly only strips a POSIX trailing separator.

This matters for GScan’s standalone library/CLI usage on Windows even though the hosted service runs Linux. Suggested fix: calculate depth from a separator-neutral representation (e.g. split on `/[\\\\/]/`, or normalize first), and use `path.dirname(matches[0])` to derive the matched directory rather than removing `index.hbs` and a literal `/`. It would be worth adding a test with Windows-style match strings.

acburdine:
Good catch — fixed in 5fb5bad. Depth is now computed on a separator-neutral split (/[\\/]/) instead of '/', and the theme-root derivation uses path.dirname()/path.join() (platform-native) instead of the manual regex strip. Extracted the comparator (compareMatchDepth) so it's unit-testable directly with Windows-style ('\\') match strings on any OS, independent of the platform-native path step.

PR #927, opened 2026-09-08
✨ Split gscan lib/cli from web app

acburdine:
no ref
- switch repo to pnpm workspaces
- move web app to `apps/web`
- move gscan lib to `packages/gscan`
- update ci/docker builds


.dockerignore
troyciesco: these live in `apps/web` now, so i think we need to update these. The same might also be true for some of  these other ones like `.env`?

packages/gscan/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "name": "Ghost Foundation",
+    "email": "hello@ghost.org",
+    "web": "https://ghost.org/"
+  },
+  "main": "lib",
+  "license": "MIT",
+  "scripts": {
+    "test": "NODE_ENV=testing vitest run --coverage",
+    "preship": "pnpm -w run test && pnpm -w run lint",
+    "ship": "pro-ship",
```
troyciesco, line 36: You def know more about this than me and i'm not sure if you're planning to change/remove this or something, but i think bc there's no `pnpm-lock.yaml` in `packages/gscan` and no `packageManager` field here, I think pro-ship will fall back to yarn instead of pnpm. see here: https://github.com/TryGhost/Pro-Packages/blob/main/packages/pro-ship/lib/pro-ship.js#L116

package.json
```diff
@@ -27,54 +18,24 @@
...
   },
-  "main": "lib",
   "license": "MIT",
   "scripts": {
-    "start": "node app/index.js",
-    "dev": "NODE_ENV=development DEBUG=gscan:* nodemon",
-    "lint": "eslint . --ext .js --cache",
-    "test": "NODE_ENV=testing vitest run --coverage",
+    "start": "pnpm --filter @tryghost/gscan-web run start",
+    "dev": "pnpm --filter @tryghost/gscan-web run dev",
```
troyciesco, line 24: i _think_ before this woulda done `nodemon` for both the webapp and the package, so if you made changes to the package you could test it by uploading to the webapp, but now you'd have to restart to pick up package changes.

troyciesco (changes requested):

package.json
acburdine: fixed 👍🏻

packages/gscan/package.json
acburdine: fixed 👍🏻

.dockerignore
acburdine: fixed 👍🏻

troyciesco (approved):
noting for posterity this will close https://github.com/TryGhost/gscan/issues/730 🎉

